PR #13, opened 2017-11-03
Road Changes

fosskers:
### TODO

- [x] Testing environment, Terraform
- [x] Bump sbt
- [x] Benchmarks
- [ ] Road changes from ORC or Features


fosskers:
https://github.com/locationtech/jts/blob/master/modules/core/src/main/java/org/locationtech/jts/algorithm/Length.java#L31

How JTS calculates Line length. I'm not sure of the units.

fosskers:
> [info] 0.0 kilometers of roads were changed.

Works like a charm. :sunglasses:

src/analytics/src/main/scala/osmesa/analytics/oneoffs/Roads.scala
```diff
@@ -0,0 +1,64 @@
...
+  }
+)
+
+object Analysis {
+
+  /** How many kilometers of road changed in all the Ways present in the given DataFrame? */
+  def roads(data: DataFrame)(implicit ss: SparkSession): Try[Double] = {
+
+    Try(osm.fromDataFrame(data)).map { case (nodes, ways, relations) =>
+      val roadsOnly: RDD[(Long, osm.Way)] = ways.filter(_._2.data.tagMap.contains("road"))
```
lossyrob: My sense is that if this filter gets performed prior to parsing the DF into case classes, it'll be more performant.

fosskers:
Oh absolutely - I was going for the more immediately available approach first. On a Friday afternoon it wasn't immediately obvious to me what the SparkSQL would've been to say:

> Take this Dataframe and give me a reduced form of it that contains all Nodes and Ways, such that the `TAGS` map in each way contains the key `road`.

src/analytics/src/main/scala/osmesa/analytics/oneoffs/Roads.scala
fosskers: Oh absolutely - I was going for the more immediately available approach first. On a Friday afternoon it wasn't immediately obvious to me what the SparkSQL would've been to say:

> Take this Dataframe and give me a reduced form of it that contains all Nodes and Ways, such that the `TAGS` map in each way contains the key `road`.

lossyrob: Gotcha, that makes sense. The SQL isnt directly obvious to me as well, might take `udf` to do it. When you get there, happy to talk through it more

fosskers:
What was needed was to search for the key `highway`. See:  https://wiki.openstreetmap.org/wiki/Key:highway

lossyrob:
Check these values too:

```scala
 val tags =
        Set(
            "motorway",
"trunk",
"motorway_link",
"trunk_link",
"primary",
"secondary",
"tertiary",
"primary_link",
"secondary_link",
"tertiary_link",
"service",
"residential",
"unclassified",
"living_street",
"road"
            )

**Comment by fosskers** [comment]
Roger. My tests revealed a bug in VectorPipe that I'm fixing now / tomorrow morning.

**Comment by fosskers** [comment]
```
[info] 199.53569223578558 kilometers of roads were changed.
```

**Comment by fosskers** [comment]
Now to produce actual kilometre distances.

**Comment by fosskers** [comment]
My use of Haversine is wrong, I'll correct it now.

File: src/analytics/src/main/scala/osmesa/analytics/oneoffs/Roads.scala [hunk 2]
```diff
@@ -65,11 +65,10 @@ object Analysis {
...
   /** How long is a Line, in metres? */
   private[this] def metres(line: Line): Double = {
-    val ps: Array[Point] = line.points // TODO Wasteful.
-    val head: Point = ps.head
-    val last: Point = ps.last
+    val ps: List[Point] = line.points.toList
+    val pairs: Iterator[(Point, Point)] = ps.iterator.zip(ps.tail.iterator)
 
-    Haversine(head.x, head.y, last.x, last.y)
+    pairs.foldLeft(0d) { case (acc, (p,c)) => acc + Haversine(p.x, p.y, c.x, c.y) }
```
mojodna: I had this as:

```scala
line.points.sliding(2).map(pair => Haversine(pair.head.x, pair.head.y, pair.last.x, pair.last.y)).reduce(_ + _)
```

Equivalent? (I'm coming up to speed on Scala, so advice helps.)

fosskers: I'll benchmark it.

fosskers:
```
[info] Benchmark                Mode  Cnt   Score   Error  Units
[info] MetresBench.iterator10   avgt    5   2.732 ± 0.119  us/op
[info] MetresBench.iterator100  avgt    5  25.007 ± 0.393  us/op
[info] MetresBench.manual10     avgt    5   2.407 ± 0.044  us/op
[info] MetresBench.manual100    avgt    5  21.972 ± 0.317  us/op
[info] MetresBench.sliding10    avgt    5   4.863 ± 0.077  us/op
[info] MetresBench.sliding100   avgt    5  48.641 ± 0.420  us/op
[info] MetresBench.while10      avgt    5   2.070 ± 0.019  us/op
[info] MetresBench.while100     avgt    5  20.619 ± 0.076  us/op
```
`sliding` kills performance, looks like. My guess is from all the allocations of intermediate `Array`s that come in as the `pair` arg.

src/analytics/src/main/scala/osmesa/analytics/oneoffs/Roads.scala
```diff
@@ -0,0 +1,96 @@
...
+  /* Lengths of new roads created by each user */
+  def newRoadsByUser(data: DataFrame)(implicit ss: SparkSession): Try[RDD[(Long, Double)]] = {
+
+    Try(osm.fromDataFrame(data)).map { case (nodes, ways, relations) =>
+      val roadsOnly: RDD[(Long, osm.Way)] =
+        ways.filter(_._2.meta.tags.get("highway").map(highways.contains(_)).getOrElse(false))
+
+      /* Roads that only had one version, implying that they were new */
+      val news: RDD[(Long, osm.Way)] = roadsOnly.groupByKey
+        .filter { case (_, iter) => iter.size === 1 }
```
mojodna: This reads to me as though it's filtering for roads that were _only_ created, not the version of the road when it was initially created. For that, I think `roadsOnly` can have this applied:

```scala
.filter(_._2.meta.version == 1)
```

fosskers: It was the original intent to get the ones that were only created, yeah. The code is simple enough that we could produce endless variations, mind you. The code here is at best an exploratory one-off.

moradology:
LGTM

Issue #11, opened 2017-10-30
Generate statistics with Zeppelin and VectorPipe directly from ORC


lossyrob:
## Per Changeset stats
- [ ] km of roads edited
- [ ] number of road edits
- [ ] number of buildings edited
- [ ] number of POIs edited (amenity)
- [ ] number of waterway edits
- [ ] km of waterways edited
- [ ] extent of edits
- [ ] countries edited in
- [ ] hashtags participated in

## Per Hashtag stats
- [ ] number of participants
- [ ] km of roads edited
- [ ] number of buildings edited
- [ ] number of POIs edited (amenity)
- [ ] km of waterways edited

fosskers:
The GT team has been having a lot of issues with Zeppelin lately. Is there value in prototyping locally first?

Also, do we have a decent history file to work off of?

moradology:
Getting a small EMR cluster up and playing around on Zeppelin (which I can assist with if you encounter trouble) seems to me the simplest approach. I'll assist with getting the Zeppelin instance running if that proves difficult

fosskers:
We can't do buildings yet, as we haven't ported over the `Line <-> Polygon` reassociation code for Planet History. Should I do that first?

kamicut:
A couple more stats would be useful to do a breakdown chart (Type of edits pie chart)

- Number of roads edited
- Number of waterways edited

fosskers:
Those should be easy to get in passing while doing the more involved ones.

mojodna:
These statistics are all calculated now.

PR #20, opened 2017-12-02
Stats calculation

lossyrob:
This PR adds a statistics calculation job, which provides the stats of #11.

This is WIP. Steps left:
- [x] Create test framework and verify current progress
- [x] Calculate countries accurately for any changed element
- [x] Calculate KM of road and waterways in the stats process
- [x] Merge with the query server work
- [x] Dump results to S3
- [x] Respond to all PR comments


src/analytics/src/main/scala/osmesa/analytics/ColumnFunctions.scala
```diff
@@ -0,0 +1,68 @@
...
+    }).apply(col).as[Boolean]
+
+  def isPOI(col: Column): TypedColumn[Any, Boolean] =
+    udf[Boolean, Map[String, String]]({ tags =>
+      tags contains "amenity"
+    }).apply(col).as[Boolean]
+
+  def isWaterway(col: Column): TypedColumn[Any, Boolean] =
+    udf[Boolean, Map[String, String]]({ tags =>
+      tags contains "waterway"
```
mojodna: Linear water features are `('river', 'canal', 'stream', 'brook', 'drain', 'ditch')` per http://wiki.openstreetmap.org/wiki/Waterways

src/analytics/src/main/scala/osmesa/analytics/ColumnFunctions.scala
```diff
@@ -0,0 +1,68 @@
...
+ * UDFs for working with OSM data in Spark DataFrames.
+ *
+ * @author sfitch
+ * @since 4/3/17
+ */
+trait ColumnFunctions {
+  private implicit def arrayEnc[T: TypeTag]: Encoder[Array[T]] = ExpressionEncoder()
+  private implicit def boolEnc: Encoder[Boolean] = ExpressionEncoder()
+
+  private val highways: Set[String] = Set(
```
mojodna: Replaced by `Constants.ROAD_VALUES`.

src/analytics/src/main/scala/osmesa/analytics/Constants.scala
```diff
@@ -0,0 +1,10 @@
+package osmesa.analytics
+
+object Constants {
+  val ROAD_VALUES =
+    Set(
+      "motorway", "trunk", "motorway_link", "trunk_link", "primary", "secondary", "tertiary",
+      "primary_link", "secondary_link", "tertiary_link", "service", "residential", "unclassified",
+      "living_street", "road"
```
mojodna: Should this include paths, footpaths, cycleways, bridleways, etc. or is this intended for "roads" only vs. navigable features?

src/analytics/src/main/scala/osmesa/analytics/stats/StatTopics.scala
```diff
@@ -0,0 +1,28 @@
...
+  def ALL = List(BUILDING, ROAD, WATERWAY, POI)
+
+  val BUILDING = "B"
+  val ROAD = "R"
+  val WATERWAY = "W"
+  val POI = "P"
+
+  def tagsToTopics(tags: Map[String, String]): Array[StatTopic] = {
+    val b = mutable.ArrayBuffer[StatTopic]()
+    tags.get("highway") match {
```
mojodna:
```
Caused by: org.apache.spark.SparkException: Failed to execute user defined function(anonfun$statTopics$1: (map<string,string>) => array<string>)
  at org.apache.spark.sql.catalyst.expressions.GeneratedClass$GeneratedIterator.processNext(Unknown Source)
  at org.apache.spark.sql.execution.BufferedRowIterator.hasNext(BufferedRowIterator.java:43)
  at org.apache.spark.sql.execution.WholeStageCodegenExec$$anonfun$8$$anon$1.hasNext(WholeStageCodegenExec.scala:395)
  at scala.collection.Iterator$$anon$11.hasNext(Iterator.scala:408)
  at scala.collection.Iterator$$anon$11.hasNext(Iterator.scala:408)
  at scala.collection.Iterator$$anon$11.hasNext(Iterator.scala:408)
  at org.apache.spark.shuffle.sort.UnsafeShuffleWriter.write(UnsafeShuffleWriter.java:166)
  at org.apache.spark.scheduler.ShuffleMapTask.runTask(ShuffleMapTask.scala:96)
  at org.apache.spark.scheduler.ShuffleMapTask.runTask(ShuffleMapTask.scala:53)
  at org.apache.spark.scheduler.Task.run(Task.scala:108)
  at org.apache.spark.executor.Executor$TaskRunner.run(Executor.scala:335)
  at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
  at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
  at java.lang.Thread.run(Thread.java:748)
Caused by: scala.MatchError: Some(cycleway) (of class scala.Some)
  at osmesa.analytics.stats.StatTopics$.tagsToTopics(StatTopics.scala:17)
  at osmesa.analytics.stats.CalculateStats$$anonfun$statTopics$1.apply(CalculateStats.scala:36)
  at osmesa.analytics.stats.CalculateStats$$anonfun$statTopics$1.apply(CalculateStats.scala:36)
  ... 14 more
```

src/analytics/src/main/scala/osmesa/analytics/ColumnFunctions.scala
lossyrob: :+1:

src/analytics/src/main/scala/osmesa/analytics/stats/StatTopics.scala
lossyrob: :+1:

mojodna:
```patch
commit 9f76d051781513a31554fd30ec6e38f290435809
Author: Seth Fitzsimmons <seth@mojodna.net>
Date:   Tue Dec 5 19:29:03 2017 -0800

    Filter changesets to more efficiently work with extracts

diff --git a/docker-compose.services.yml b/docker-compose.services.yml
index ce1360b..e769bcb 100644
--- a/docker-compose.services.yml
+++ b/docker-compose.services.yml
@@ -1,5 +1,11 @@
 version: '3'
 services:
+  base:
+    image: quay.io/azavea/osmesa-base:latest
+    build:
+      context: ./services/services-base
+      dockerfile: Dockerfile
+    command: ["true"]
   hdfs-name:
     image: quay.io/azavea/osmesa-hdfs:latest
     build:
diff --git a/src/analytics/src/main/scala/osmesa/analytics/stats/CalculateStats.scala b/src/analytics/src/main/scala/osmesa/analytics/stats/CalculateStats.scala
index f9004d4..820582f 100644
--- a/src/analytics/src/main/scala/osmesa/analytics/stats/CalculateStats.scala
+++ b/src/analytics/src/main/scala/osmesa/analytics/stats/CalculateStats.scala
@@ -56,10 +56,15 @@ object CalculateStats {
           }).toList.map { case (k, v) => (k, new UpstreamTopics(v)) }
         }
 
+    // inner join changesets against history to include only changesets referenced by the available data
+    val filteredChangesets = history
+      .select('changeset.as("id"))
+      .distinct
+      .join(changesets.select('id, 'uid, 'user, 'created_at, 'closed_at, 'tags), Seq("id"))
 
     // Create the base set of ChangesetStats we'll be joining against
     val initialChangesetStats =
-      changesets.
+      filteredChangesets.
         select($"id", $"uid", $"user", $"created_at", $"closed_at", $"tags", hashtags($"tags").alias("hashtags")).
         map { row =>
           val changeset = row.getAs[Long]("id")
```

mojodna:
```patch
commit 63c585819b7c0e86a1276acdde044e0cc71e6263
Author: Seth Fitzsimmons <seth@mojodna.net>
Date:   Tue Dec 5 19:40:44 2017 -0800

    Improved waterway + building checks

diff --git a/src/analytics/src/main/scala/osmesa/analytics/stats/StatTopics.scala b/src/analytics/src/main/scala/osmesa/analytics/stats/StatTopics.scala
index 43bb195..1091294 100644
--- a/src/analytics/src/main/scala/osmesa/analytics/stats/StatTopics.scala
+++ b/src/analytics/src/main/scala/osmesa/analytics/stats/StatTopics.scala
@@ -19,8 +19,16 @@ object StatTopics {
       case _ => ()
     }
 
-    if(tags contains "building") { b += StatTopics.BUILDING }
-    if(tags contains "waterway") { b += StatTopics.WATERWAY }
+    tags.get("waterway") match {
+      case Some(v) if Constants.WATERWAY_VALUES.contains(v) => b += StatTopics.WATERWAY
+      case _ => ()
+    }
+
+    tags.get("building") match {
+      case Some(v) if v.toLowerCase != "no" => b += StatTopics.BUILDING
+      case _ => ()
+    }
+
     if(tags contains "amenity") { b += StatTopics.POI }
 
     b.toArray
```

mojodna:
Fun fact: `building=no` occurs 13,148 times, presumably to mark facilities that would otherwise be implicitly assumed to be buildings (e.g. schools).

mojodna:
```patch
commit dda5260857cb13417c8dd7f9d1bcf61e39396a0f
Author: Seth Fitzsimmons <seth@mojodna.net>
Date:   Tue Dec 5 20:20:59 2017 -0800

    Add isExtract to Options

diff --git a/src/analytics/src/main/scala/osmesa/analytics/stats/CalculateStats.scala b/src/analytics/src/main/scala/osmesa/analytics/stats/CalculateStats.scala
index 820582f..82e4328 100644
--- a/src/analytics/src/main/scala/osmesa/analytics/stats/CalculateStats.scala
+++ b/src/analytics/src/main/scala/osmesa/analytics/stats/CalculateStats.scala
@@ -26,7 +26,8 @@ object CalculateStats {
   case class Options(
     changesetPartitionCount: Int = 1000,
     wayPartitionCount: Int = 1000,
-    nodePartitionCount: Int = 10000
+    nodePartitionCount: Int = 10000,
+    isExtract: Boolean = false
   )
 
   object Options {
@@ -56,11 +57,19 @@ object CalculateStats {
           }).toList.map { case (k, v) => (k, new UpstreamTopics(v)) }
         }
 
-    // inner join changesets against history to include only changesets referenced by the available data
-    val filteredChangesets = history
-      .select('changeset.as("id"))
-      .distinct
-      .join(changesets.select('id, 'uid, 'user, 'created_at, 'closed_at, 'tags), Seq("id"))
+    val filteredChangesets = options match {
+      case Options(_, _, _, v) if v => {
+        // inner join changesets against history to include only changesets referenced by the available data
+        history
+          .select('changeset.as("id"))
+          .distinct
+          .join(changesets.select('id, 'uid, 'user, 'created_at, 'closed_at, 'tags), Seq("id"))
+      }
+      case _ =>  {
+        changesets
+          .select('id, 'uid, 'user, 'created_at, 'closed_at, 'tags)
+      }
+    }
 
     // Create the base set of ChangesetStats we'll be joining against
     val initialChangesetStats =
```

src/analytics/src/main/scala/osmesa/analytics/stats/CalculateStats.scala
```diff
@@ -0,0 +1,365 @@
...
+        }.
+        rdd.
+        map { case (changeset, userId, userName, createdAt, closedAt, editor, hashtags) =>
+          val stats =
+            ChangesetStats(
+              changeset,
+              userId,
+              userName,
+              createdAt.toInstant,
+              closedAt.toInstant,
```
mojodna: I've noticed the use of `Instant`s elsewhere; is there a particular reason to use them over `Timestamp` (other than `java.time` seeming more general than `java.sql`.

I ask because converting these back into `DataFrame`s (for the purpose of writing them out) fails due to a lack of `java.time.Instant` encoder.

```scala
val statsDF = changesetStats.values.toDF("changeset", "uid", "username", "created_at", "closed_at", "editor", "hashtags", "roads_added", "roads_modified", "buildings_added", "buildings_modified", "waterways_added", "waterways_modified", "pois_added", "pois_modified", "km_road_added", "km_road_modified", "km_waterway_added", "km_waterway_modified", "countries")
statDF.repartition(1).write.format("json").save("stats")
```

(The above works after leaving them as `Timestamp`s.)

src/analytics/src/main/scala/osmesa/analytics/Constants.scala
lossyrob: We're sticking with roads for now, but we should refactor to work with arbitrary topics at a later point.

src/analytics/src/main/scala/osmesa/analytics/stats/CalculateStats.scala
lossyrob: The domain models that Nathan was working on had Instants. Definitely not tied to the type; some of that might be fleshed out when we combine this PR with the query server PR to get on the same case classes.

mojodna: (Don't consider this a problem; I've figured out how to map them back to `Timestamp`s before converting to `DataFrame`s.)

src/stats/build.sbt
```diff
@@ -0,0 +1,81 @@
...
+  "com.typesafe.akka"           %% "akka-http-testkit"    % "10.0.10",
+  "ch.megard"                   %% "akka-http-cors"       % "0.2.2",
+  "io.circe"                    %% "circe-core"           % "0.9.0-M2",
+  "io.circe"                    %% "circe-generic"        % "0.9.0-M2",
+  "io.circe"                    %% "circe-generic-extras" % "0.9.0-M2",
+  "io.circe"                    %% "circe-parser"         % "0.9.0-M2",
+  "io.circe"                    %% "circe-optics"         % "0.9.0-M2",
+  "de.heikoseeberger"           %% "akka-http-circe"      % "1.17.0",
+  decline,
+  hive % "provided",
```
mojodna: `sparkHive`

lossyrob: fixed.

src/analytics/build.sbt
```diff
@@ -6,14 +6,14 @@ dependencyOverrides += "com.fasterxml.jackson.core" % "jackson-core" % "2.6.7"
...
-
 libraryDependencies ++= Seq(
+  // Deal with GeoMesa dependency that breaks Spark 2.2
+  "org.json4s" %% "json4s-native" % "3.2.11",
+  "org.json4s" %% "json4s-core" % "3.2.11",
+  "org.json4s" %% "json4s-ast" % "3.2.11",
+
   decline,
-  hive % "provided",
+  sparkHive,// % "provided",
```
mojodna: dropping `% provided` breaks the ability to run this on a cluster.

lossyrob: true, sorry shouldn't have committed that

src/analytics/src/main/scala/osmesa/analytics/stats/CalculateStats.scala
```diff
@@ -92,20 +101,20 @@ object CalculateStats {
...
         where("type == 'relation'").
         select(
           $"id",
           $"changeset",
           $"version",
           $"members",
-          statTopics($"tags").as("statTopics")
-        ).
-        where(size($"statTopics") > 0)
+          statTopics($"tags", "relation").as("statTopics")
```
mojodna: `$("type")` would work here and elsewhere and reduce the risk of copypasta.

src/analytics/src/main/scala/osmesa/analytics/stats/StatTopics.scala
```diff
@@ -8,20 +8,28 @@ object StatTopics {
...
-      case Some(v) if Constants.WATERWAY_VALUES.contains(v) => b += StatTopics.WATERWAY
-      case _ => ()
+    if(osmType == "way") {
+      tags.get("waterway") match {
+        case Some(v) if Constants.WATERWAY_VALUES.contains(v) => b += StatTopics.WATERWAY
+        case _ => ()
+      }
     }
 
     tags.get("building") match {
```
mojodna: Also `way`-specific.

mojodna: (`amenity` should probably be restricted to `way` and `node`)

src/analytics/src/main/scala/osmesa/analytics/stats/WayLengthCalculator.scala
```diff
@@ -11,6 +11,14 @@ object WayLengthCalculator {
     *
     * instant = milliseconds from epoch
     *
+    * Computes the length differences based on changests. This means if there
+    * are more than one version of the way in a single changeset, this will
+    * only consider the second way and the difference of length between
```
mojodna: second → last

src/analytics/src/main/scala/osmesa/analytics/stats/WayLengthCalculator.scala
```diff
@@ -11,6 +11,14 @@ object WayLengthCalculator {
     *
     * instant = milliseconds from epoch
     *
+    * Computes the length differences based on changests. This means if there
```
mojodna: *changesets

src/analytics/src/main/scala/osmesa/analytics/stats/WayLengthCalculator.scala
```diff
@@ -37,9 +45,38 @@ object WayLengthCalculator {
...
+      }
+
+      if(idx < 0) { None }
+      else { Some(waysArr(idx)) }
+    }
+
+    /** Gets the length of this way at an instant.
+      *  To be safe and not blow up the job on bad data, if there are any
+      * nodes the way is pointing to that we don't have in `nodes`, just skip
+      * this version of the way.
```
mojodna: Should we dump these somehow? Maybe even just as a count (per-changeset, to help track them down).

lossyrob: Would be good to, we should have some sort of logging that writes out bad geoms. That can be a TODO

src/analytics/src/main/scala/osmesa/analytics/stats/CalculateStats.scala
```diff
@@ -0,0 +1,486 @@
...
+          $"nds.ref".as("nodes"),
+          statTopics($"tags", "way").as("statTopics")
+        ).
+        rdd.
+        map { row =>
+          val id = row.getAs[Long]("id")
+          val changeset = row.getAs[Long]("changeset")
+          val version = row.getAs[Long]("version")
+          val instant = row.getAs[Timestamp]("timestamp").toInstant.toEpochMilli
+          val nodeIds = row.getAs[Seq[Long]]("nodes").toArray
```
mojodna: @lossyrob this is one of 2 spots that I suspect may be blowing up Kryo (when there are lots of nodes associated with a way). This is bypassed in the relation handling by using `explode` to create individual rows.

src/analytics/src/main/scala/osmesa/analytics/stats/CalculateStats.scala
```diff
@@ -0,0 +1,486 @@
...
+
+    /** Gather the ways from the perspective of each node.
+      * Grab all ways, not just ways with relevant topics,
+      * since we need to know country edits
+      */
+    val waysForNodes: RDD[(Long, List[(WayId, UpstreamInfoMap)])] =
+      generateUpstreamInfoMaps {
+        wayInfo.
+          flatMap { case (wayId, (changeset, version, instant, nodeIds, topics)) =>
+            nodeIds.map { nodeId =>
```
mojodna: This is the other (but is related to `wayInfo`).

src/analytics/src/main/scala/osmesa/analytics/stats/CalculateStats.scala
mojodna: The largest way has 49,189 refs.

(The most tags a single element has is 7,838. The most members a relation has is 69,643.)

src/analytics/src/main/scala/osmesa/analytics/stats/CalculateStats.scala
ayushhooda: @mojodna Can you describe how to convert Instant fields back to Dataframes

mojodna: First, convert them to `Timestamp`s (since Spark knows how to serialize them):

```scala
import java.time.Instant
import java.sql.Timestamp

new Timestamp(Instant.now.toEpochMilli)
```

Then, `toDF` should just work.

PR #23, opened 2017-12-11
Move models to common

moradology:
This PR breaks models out to a common project


src/analytics/src/main/scala/osmesa/analytics/OSMOrc.scala
```diff
@@ -1,14 +0,0 @@
-package osmesa.analytics
```
lossyrob: add back

src/analytics/src/main/scala/osmesa/analytics/oneoffs/FootprintByCampaign.scala
```diff
@@ -225,6 +225,13 @@ object FootprintByCampaign {
         val n =  URLEncoder.encode(name, "UTF-8")
         s"s3://${bucket}/${prefix}/hashtag/${n}/${zoom}/${sk.col}/${sk.row}.mvt"
       }
+  def save(zoom: Int, vectorTiles: RDD[((SpatialKey, String), VectorTile)]) = {
```
lossyrob: remove func

PR #27, opened 2018-02-13
Import DataFrame-based code for ingesting OSM

jpolchlo:
This PR pulls in code contributed by Seth Fitzsimmons into the Ingest project as a means to replace some Vectorpipe functions that have not scaled well to planet-scale ingests.  These functions have been confirmed to run on a subset of OSM, but not on the full planet history file.

### DIFF
--- a/src/ingest/src/main/scala/osmesa/ingest/ProcessOSM.scala
+++ b/src/ingest/src/main/scala/osmesa/ingest/ProcessOSM.scala
@@ -1,0 +1,395 @@
+package osmesa
+
+import com.vividsolutions.jts.geom.Coordinate
+import geotrellis.proj4.{LatLng, WebMercator}
+import geotrellis.vector._
+import geotrellis.vector.io._
+import geotrellis.vector.io.json._
+import org.apache.spark.sql._
+import org.apache.spark.sql.expressions.Window
+import org.apache.spark.sql.functions._
+import spray.json._
+
+import java.io._
+
+object ProcessOSM {
+
+  def preprocessNodes(history: DataFrame): DataFrame = {
+    import history.sparkSession.implicits._
+
+    // Convert BigDecimals to double
+    // Reduces size taken for representation at the expense of some precision loss.
+    val double = udf((bd: java.math.BigDecimal) => {
+      Option(bd).map(_.doubleValue).getOrElse(Double.NaN)
+    })
+
+    // Associate last-available tags to deleted nodes and clean coordinates
+    @transient val idByUpdated = Window.partitionBy('id).orderBy('version)
+
+    // when a node has been deleted, it doesn't include any tags; use a window function to retrieve the last tags present and use those
+    // this is suitable for appending to directly (since none of the values need to change ever)
+    val nodes = history
+      .where('type === "node")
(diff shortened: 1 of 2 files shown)


src/ingest/src/main/scala/osmesa/ingest/ProcessOSM.scala
```diff
@@ -0,0 +1,399 @@
...
+    object Resource {
+      def apply(name: String): String = {
+        val stream: InputStream = getClass.getResourceAsStream(s"/$name")
+        try { scala.io.Source.fromInputStream( stream ).getLines.mkString(" ") } finally { stream.close() }
+      }
+    }
+
+    case class CountryId(code: String)
+
+    object MyJsonProtocol extends DefaultJsonProtocol {
```
mojodna: There's another way to do this in https://github.com/azavea/osmesa/blob/e84e37be91535607cb9b0d6758375cdb7263abe6/src/analytics/src/main/scala/osmesa/analytics/Countries.scala#L13-L29

It's done this way here because the `Countries.scala` approach didn't work w/ Zeppelin, as fragments are effectively `:paste`d into the interpreter.

src/ingest/src/main/scala/osmesa/ingest/ProcessOSM.scala
```diff
@@ -180,7 +184,7 @@ object ProcessOSM {
       .distinct
 
     val fullReferencedWays = allReferencedWays
-      .select('changeset, 'id, 'version, 'updated, 'visible, posexplode_outer('nds).as(Seq("idx", "ref")))
+      .select('changeset, 'id, 'version, 'updated, 'visible, posexplode('nds).as(Seq("idx", "ref")))
```
mojodna: This actually does need to be `posexplode_outer` (though that does force a dependency on Spark-2.2.0+); without it, deleted (`visible == false`) elements expand to 0 rows (vs. 1 matched w/ `null`s), removing them from subsequent processing (and causing the eventual `validUntil` to become invalid, since the set of reconstructed geometries will only include visible versions).

jpolchlo: Very good.  The spark documentation that I found didn't clearly declare the difference between the two versions.  I'll fix this.

src/ingest/src/main/scala/osmesa/ingest/ProcessOSM.scala
jpolchlo: I'm supposing, based on your commenting, that you'd prefer to use the above?

jpolchlo:
And sorry for the o'erhasty merge, @mojodna.  Didn't think to ask for your review.  Will ask in the future!

src/ingest/src/main/scala/osmesa/ingest/ProcessOSM.scala
mojodna: No preference; I was cargo culting at this point to get it working. Mainly flagging that there was no good reason for it being this way.

mojodna:
No problem!

Looking forward to using `spark-submit` directly ;-)

PR #32, opened 2018-02-15
Initial vector tile generation

jpolchlo:
This PR represents the current progress of this tool toward being able to produce a catalog of vector tiles.  We have produced some vector tiles using this feature branch, but cannot guarantee correctness of the results.  Larger-scale runs will be necessary to extend our capabilities in this regard, but this PR represents our current status.


src/ingest/src/main/scala/osmesa/ingest/GenerateVT.scala
```diff
@@ -0,0 +1,102 @@
...
+      val (pts, ls, mls, ps, mps) = tup
+
+    }
+
+    keyedGeoms
+      .combineByKey(create, merge, combine)
+      .map(
+  }
+
+  private def geometryToGridCells(g: Geometry, re: RasterExtent): Seq[SpatialKey] = {
```
mojodna: This is the equivalent of `clipToGrid` (without the clipping)?

src/ingest/src/main/scala/osmesa/ingest/GenerateVT.scala
```diff
@@ -0,0 +1,74 @@
...
+      (pt1 ++ pt2, l1 ++ l2, ml1 ++ ml2, p1 ++ p2, mp1 ++ mp2)
+    }
+
+    keyedGeoms
+      .combineByKey(create, merge, combine)
+      .map { case (sk, tup) => {
+        val (pts, ls, mls, ps, mps) = tup
+        val extent = layout.mapTransform(sk)
+
+        val layer = StrictLayer(
```
mojodna: Does this clip geometries to tile bounds?

src/ingest/src/main/scala/osmesa/ingest/GenerateVT.scala
jpolchlo: Yes. (I don't understand why this diff is out of date.  I have an updated version of this that is cleaner and uses an existing GT feature to achieve this same end.)

src/ingest/src/main/scala/osmesa/ingest/GenerateVT.scala
jpolchlo: (Sorry if this is a dup, something was weird with github, and my last comment has vanished)

Yes.  This will quickly assign spatial keys to geometries, but will not clip to the tile.

jpolchlo: Specifically, line 28
```scala
val keys = layout.mapTransform.keysForGeometry(g)
```
provides the keys.

src/ingest/src/main/scala/osmesa/ingest/GenerateVT.scala
mojodna: I was commenting directly on a commit rather than the full diff, sorry!

src/ingest/src/main/scala/osmesa/ingest/GenerateVT.scala
mojodna: Gotcha.

Does `StrictLayer` do any clipping when producing the MVT layer? (If not, my concern is that individual tiles will be **much** heavier than they need to be, including geometry components outside the extent of a given tile.)

A lesser concern (to potentially be borne out in testing, so maybe this is moot) is that duplicating full geometries for each key (esp. for large geometries) and then grouping by key will produce large, unbalanced partitions and more shuffle than is necessary.

mojodna: It depends. Entire geometries that expand far beyond the tile boundaries (not sure how far; values < 0 or >= 4096 are "invisible" but considered when rendering; this may be how feature rendering is isolated to the visible contents of a tile) are valid (if inefficient).

For purely analytical purposes, the 2 obvious options seem like a) full geometries or b) geometries clipped to tile bounds.

For cartographic purposes (typically labels / markers), Mapnik (the original reference implementation) supports per-layer buffers (measured in "pixels") that are used to expand a tile's boundaries and clip features (no reason other than convenience that they can't be per-feature buffers). When layers are buffered, features will be duplicated between tiles and considered when doing collision / duplicate label detection during the render step (I'm mainly familiar with raster rendering). For labeled points that are "off-screen" (relative to a tile being rendered), the label can be rasterized partially off-canvas and it will line up with a neighboring label that has been cropped.

(If my assumption about "invisible coordinates" holds, it may be possible "clip" geometries during the command generation process by dropping coordinates out of the stream when they're outside the buffered tile boundary (in screen coordinates), except when crossing one of the corners.)

mojodna: I'd forgotten about the 3x3 clipping done by Mapbox + Mapzen, referenced in https://github.com/geotrellis/vectorpipe/issues/11.

jpolchlo:
This most recent commit adds clipping to the 3x3 spatial key neighborhood.  The visualized results are much clearer.  Performance isn't crazily different compared to the unclipped version—at least on a small input.  I've also added a 500 millisecond timeout on the intersection process to ensure that any wacky geometries that cause problems don't ham up the works completely.  These failing geometries will trigger a log message giving both the geometry and the extent we were trying to clip to (this will be the 3x3 extent).

src/ingest/src/main/scala/osmesa/ingest/GenerateVT.scala
```diff
@@ -0,0 +1,164 @@
...
+      val g = feat.geom
+      val keys = layout.mapTransform.keysForGeometry(g)
+      keys.map{ k => (k, (k, feat)) }
+    }
+
+    type FeatureTup = (Seq[VTF[Point]], Seq[VTF[MultiPoint]], Seq[VTF[Line]], Seq[VTF[MultiLine]], Seq[VTF[Polygon]], Seq[VTF[MultiPolygon]])
+
+    def timedIntersect[G <: Geometry](geom: G, ex: Extent) = {
+      val future = Future { geom.intersection(ex) }
+      Try(Await.result(future, 500 milliseconds)) match {
```
mojodna: ❤️

src/ingest/src/main/scala/osmesa/ingest/GenerateVT.scala
```diff
@@ -0,0 +1,164 @@
...
+        case Success(res) => res
+        case Failure(_) =>
+          logger.warn(s"Could not intersect $geom with $ex")
```
mojodna: If we could also log the feature id, that would be awesome, since we could then track down weird OSM geometries and figure out how to deal with them.

mojodna:
Fantastic! I'm looking forward to having time to give this a shot.

jpolchlo:
Added some basic pyramiding functionality.  I don't recommend running this.  I had some stack overflow issues during some runs on a small data set.  There are only slight benefits to running this code, since the size of the data generally does not shrink as we pyramid.  Would be better to allow caching of the initial processing of the OSM results to orc files?  Or something?

jpolchlo:
Added some basic pyramiding functionality.  I don't recommend running this.  I had some stack overflow issues during some runs on a small data set.  There are only slight benefits to running this code, since the size of the data generally does not shrink as we pyramid.  Would be better to allow caching of the initial processing of the OSM results to orc files?  Or something?

mojodna:
I don't think it's in this PR directly, but vectorpipe@0.2.2 hasn't been published yet, so it needs to be `publishLocal`'d prior to successful building of the ingest assembly.

mojodna:
When producing tiles for analysis purposes (i.e. full geometries), it seems to me that generating multiple zoom levels isn't all that useful. In practice, we'd want to generate the lowest zoom for which individual tile sizes are reasonable (for generating or consuming) and screen coordinates aren't lossy (e.g. the default `tileWidth` (4096) is good for `<zoom> + 3` @ retina resolution (`512 * 2**3`)). To produce output at higher zooms, over-zooming (read subsets of individual tiles) works just as well as reading higher zoom tiles (with the precision caveat).

Generating lower zoom tiles will be necessarily lossy, as we'll want to simplify lines + polygons and selectively drop points in order to keep the tiles to a reasonable file size.

(I'm on the fence about how geometries should be clipped for analysis tiles (and can't remember off-hand how the Mapbox QA tiles are clipped.)

When generating vector tiles for cartographic purposes (i.e. w/ the OpenMapTiles or Mapzen schemas), pyramiding *definitely* makes sense, although the rules for simplifying + dropping features should be explicitly curated (i.e. @ zoom 10, only show points w/ `place=*`, lines w/ `highway={motorway,primary}`, buildings (`building=*`) larger than `x` m², ...). (Rules for clipping buffers too, I think).

> Would be better to allow caching of the initial processing of the OSM results to orc files?

I think so; I've been storing the node + way geoms on S3 separately and running processing jobs against them (e.g. region stats + changeset stats).

jpolchlo:
I'm going to merge these changes in a while if there are no objections.

jpolchlo:
Oh, and in response to your comments, @mojodna, it appears that any pyramiding work should take place in another PR since it's surely a nontrivial feature.  But I copy; over and out.

ARolek:
This PR was brought to my attention to see if any of the work we have done on the tegola project could be of assistance. My understanding is that the goal is to generate vector tiles at full fidelity for analysis purposes. I agree with @mojodna that maybe generating multiple zooms is not necessary for this use case and simplification should probably be avoided. With tegola we have dealt with a lot of simplification issues which result in invalid polygons / multi polygons which are difficult to fix. 

The main advantages I see to using a tiling approach for analysis is filtering features by zoom and then "querying" an area to stitch back together for comparison against another dataset of the same area. The algos for clipping, scaling, re-projection and fixing invalid polygons vary for different implementations so the same processing should be applied to both datasets. For tegola, we have a seeding tool and you can turn off simplification if that would be of any help for cross referencing some of this work.

Also worth noting is that MVT tiles have a buffer that needs to be considered when stitching tiles back together. Speaking of which, what's the strategy for tile stitching prior to comparison?

PR #40, opened 2018-02-21
Allow filtering to metro area polygons by country

jpolchlo:
Fixes reading of metro area shapefile and provides new command line options for filtering to metro areas in a country.

Signed-off-by: jpolchlo <jpolchlopek@azavea.com>


jpolchlo:
This feature is no longer needed as we decided to generate vector tiles for the whole planet, and not just metro areas.

PR #41, opened 2018-02-26
Improved ORC caching

jpolchlo:
Closes #39 

This PR provides ORC caching facilities.  These have been tested at scale and have been found to help.

### DIFF
--- a/deployment/makefiles/Makefile
+++ b/deployment/makefiles/Makefile
@@ -1,0 +1,134 @@
+include config-aws.mk # Vars related to AWS credentials and services used
+include config-emr.mk # Vars related to type and size of EMR cluster
+include config-run.mk # Vars related to ingest step and spark parameters
+
+ZOOM ?= 15
+
+INGEST_SRC_DIR := ../../src/ingest
+INGEST_JAR := osmesa-ingest.jar
+INGEST_ASSEMBLY := ${INGEST_SRC_DIR}/target/scala-2.11/${INGEST_JAR}
+
+ifeq ($(USE_SPOT),true)
+MASTER_BID_PRICE:=BidPrice=${MASTER_PRICE},
+WORKER_BID_PRICE:=BidPrice=${WORKER_PRICE},
+endif
+
+ifdef COLOR
+COLOR_TAG=--tags Color=${COLOR}
+endif
+
+ifndef CLUSTER_ID
+CLUSTER_ID=$(shell if [ -e "cluster-id.txt" ]; then cat cluster-id.txt; fi)
+endif
+
+rwildcard=$(foreach d,$(wildcard $1*),$(call rwildcard,$d/,$2) $(filter $(subst *,%,$2),$d))
+
+${INGEST_ASSEMBLY}: $(call rwildcard, ${INGEST_SRC_DIR}/src, *.scala) ${INGEST_SRC_DIR}/build.sbt
+	cd ${INGEST_SRC_DIR}/.. && ./sbt "project ingest" assembly
+	@touch -m ${INGEST_ASSEMBLY}
+
+create-cluster:
+	aws emr create-cluster --name "${NAME}" ${COLOR_TAG} \
+--release-label emr-5.8.0 \
+--output text \
+--use-default-roles \
+--configurations "file://$(CURDIR)/scripts/configurations.json" \
+--log-uri ${S3_URI}/logs \
+--ec2-attributes KeyName=${EC2_KEY},SubnetId=${SUBNET_ID} \
+--applications Name=Ganglia Name=Hadoop Name=Hue Name=Spark Name=Zeppelin \
+--instance-groups \
+'Name=Master,${MASTER_BID_PRICE}InstanceCount=1,InstanceGroupType=MASTER,InstanceType=${MASTER_INSTANCE}' \
+'Name=Workers,${WORKER_BID_PRICE}InstanceCount=${WORKER_COUNT},InstanceGroupType=CORE,InstanceType=${WORKER_INSTANCE}' \
+| tee cluster-id.txt
+
+upload-code: ${INGEST_ASSEMBLY}
+	@aws s3 cp ${INGEST_ASSEMBLY} ${S3_URI}/
+
+planet-ingest:
+	aws emr add-steps --output text --cluster-id ${CLUSTER_ID} \
+--steps Type=CUSTOM_JAR,Name="Planet ingest - zoom ${ZOOM}",Jar=command-runner.jar,Args=[\
+spark-submit,--master,yarn,--deploy-mode,cluster,\
+--class,osmesa.IngestApp,\
+--driver-memory,${DRIVER_MEMORY},\
+--driver-cores,${DRIVER_CORES},\
+--executor-memory,${EXECUTOR_MEMORY},\
+--executor-cores,${EXECUTOR_CORES},\
+--conf,spark.driver.maxResultSize=3g,\
+--conf,spark.dynamicAllocation.enabled=true,\
+--conf,spark.yarn.executor.memoryOverhead=${YARN_OVERHEAD},\
+--conf,spark.yarn.driver.memoryOverhead=${YARN_OVERHEAD},\
+--conf,spark.shuffle.service.por=7338,\
+--conf,spark.shuffle.io.numConnectionsPerPeer=4,\
+--conf,spark.shuffle.io.connectionTimeout=18000s,\
+--conf,spark.shuffle.service.enabled=true,\
+${S3_URI}/${INGEST_JAR},\
+--orc=${PLANET_ORC},\
+--bucket=${S3_BUCKET},\
+--key=vectortiles,\
+--layer=history,\
+--cache=${ORC_CACHE_LOCATION},\
+--zoom=${ZOOM}\
+] | cut -f2 | tee last-step-id.txt
+
+kenya-ingest:
+	aws emr add-steps --output text --cluster-id ${CLUSTER_ID} \
+--steps Type=CUSTOM_JAR,Name="Kenya ingest - zoom ${ZOOM}",Jar=command-runner.jar,Args=[\
+spark-submit,--master,yarn,--deploy-mode,cluster,\
+--class,osmesa.IngestApp,\
+--driver-memory,${DRIVER_MEMORY},\
+--driver-cores,${DRIVER_CORES},\
+--executor-memory,${EXECUTOR_MEMORY},\
+--executor-cores,${EXECUTOR_CORES},\
+--conf,spark.driver.maxResultSize=3g,\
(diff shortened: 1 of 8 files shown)


src/ingest/src/main/scala/osmesa/ingest/util/Caching.scala
```diff
@@ -0,0 +1,71 @@
...
+  def orc(filename: String)(sparkjob: => DataFrame)(implicit ss: SparkSession): DataFrame = {
+    logger.warn("in cache func")
+    logger.warn("bucket: " ++ bucket)
+    logger.warn(prefix(filename))
+    logger.warn(fileUri(filename))
+    logger.warn("s3exists: " ++ s3exists(filename).toString)
+    if (s3exists(filename)) {
+      ss.read.orc(fileUri(filename)).repartition(10000)
+    } else {
+      sparkjob.repartition(1).write.format("orc").save(fileUri(filename))
```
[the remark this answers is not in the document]
jpolchlo: If you mean the `repartition(1)`, I fully agree.  ORC writes were taking 20+ minutes for North America.  Given the number of times we cache, this adds up in a significant way.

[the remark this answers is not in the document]
mojodna: Eventually (and generally), it would be nice to figure out a heuristic for determining the number of target partitions. Too few and writes take forever. Too many and initializing a remote datasource takes a long time (but only on the driver, and only when doing the initial "read"). I've been using an upper limit of 100, fwiw.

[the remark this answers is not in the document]
jpolchlo: Yes.  This.  These repartitions are a bad idea.  I surely did encounter issues with these configs.  I ended up eliminating the repartition on 37, and changing the repartition on 39 to be the same as the number of worker nodes.  Cleared up many problems.  Misunderstanding the degree of participation of the master node in coordinating tasks led to some mysterious behaviors.  Getting fast performance out of ORC would seem like an easy thing (considering that the phrase "High performance" is one of the first things you see when you search for `Apache ORC`), but it's clearly non-trivial.  I think ORC writes of planet-scale cache data were taking about 20 minutes each?  I've no context for knowing if that's unreasonable.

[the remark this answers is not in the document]
mojodna: ¯\_(ツ)_/¯

It seems moderately reasonable for ~300GB of writes to S3, but the ability to parallelize is nice. More partitions possibly means more splits when reading (but also more metadata requests when initializing).

I'm really looking forward to seeing if the ORC improvements (`spark.sql.orc.impl=native`) in Spark 2.3.0 speed things up...

Using instance types that support the full 25Gbps _may_ help now that S3 supports those speeds as well.

deployment/makefiles/scripts/configurations.json
```diff
@@ -0,0 +1,88 @@
...
+  {
+    "Classification": "spark-defaults",
+    "Properties": {
+        "spark.driver.maxResultSize": "3G",
+        "spark.dynamicAllocation.enabled": "true",
+        "spark.shuffle.service.enabled": "true",
+        "spark.shuffle.compress": "true",
+        "spark.shuffle.spill.compress": "true",
+        "spark.rdd.compress": "true",
+        "spark.yarn.executor.memoryOverhead": "1G",
```
mojodna: FWIW, I've needed to use 2G when doing planet-scale processing, but I'm also not using any GC options.

(Is this intended to provide defaults, or will these always be provided in the EMR jobs triggered via the `Makefile`?)

deployment/makefiles/config-emr.mk
```diff
@@ -0,0 +1,20 @@
...
+export MASTER_INSTANCE:=m3.2xlarge
+export MASTER_PRICE := 0.13
+export DRIVER_MEMORY := 20000M
+export DRIVER_CORES := 8
+
+export WORKER_INSTANCE:=m3.xlarge
+export WORKER_PRICE := 0.13
+export WORKER_COUNT := 200
+export EXECUTOR_MEMORY := 21500M
+export EXECUTOR_CORES := 8
```
mojodna: Have you seen a benefit to providing these vs. allowing EMR to set these automatically?

deployment/makefiles/Makefile
```diff
@@ -0,0 +1,152 @@
...
+--class,osmesa.IngestApp,\
+--driver-memory,${DRIVER_MEMORY},\
+--driver-cores,${DRIVER_CORES},\
+--conf,spark.driver.maxResultSize=0,\
+--conf,spark.dynamicAllocation.enabled=true,\
+--conf,spark.yarn.executor.memoryOverhead=${YARN_OVERHEAD},\
+--conf,spark.yarn.driver.memoryOverhead=${YARN_OVERHEAD},\
+--conf,spark.shuffle.io.numConnectionsPerPeer=4,\
+--conf,spark.shuffle.io.connectionTimeout=18000s,\
+--conf,spark.shuffle.blockTransferService=nio,\
```
mojodna: What does this do? (It's missing from https://spark.apache.org/docs/latest/configuration.html for some reason)

deployment/makefiles/scripts/configurations.json
```diff
@@ -0,0 +1,88 @@
...
+        "spark.yarn.driver.memoryOverhead": "1G",
+        "spark.driver.maxResultSize": "3G",
+        "spark.executor.extraJavaOptions" : "-XX:+UseParallelGC -Dgeotrellis.s3.threads.rdd.write=64"
```
mojodna: What's the default for `geotrellis.s3.threads.rdd.write`? (I'm assuming that this speeds up vector tile writes but doesn't impact ORC writes.)

src/ingest/src/main/scala/osmesa/ingest/Ingest.scala
```diff
@@ -122,21 +126,45 @@ object IngestApp extends CommandApp(
...
-      val wayGeoms = ProcessOSM.reconstructWayGeometries(ppnodes, ppways)
-      println("PRIOR TO WAY/NODE UNION")
-      nodeGeoms.withColumn("minorVersion", lit(null).cast(IntegerType)).printSchema
-      wayGeoms.printSchema
+      println("caching...")
+      val cache = Option(new URI(cacheDir).getScheme) match {
+        case Some("file") => Caching.onFs(cacheDir)
+        case Some("s3") => Caching.onS3(cacheDir)
+        case _ => Caching.none
+      }
```
mojodna: See https://github.com/azavea/osmesa/pull/47/files#diff-9399197c9828d222bc4d93480ee442f8 for a fix for local FS caching.

src/ingest/src/main/scala/osmesa/ingest/Ingest.scala
```diff
@@ -122,21 +126,45 @@ object IngestApp extends CommandApp(
...
+
+      val orcFileRepr = orc.split("/").last.split('.').head
+
+      val clippedWorldExtent = Extent(-180, -75, 180, 85).reproject(LatLng, WebMercator)
```
mojodna: -75?

jpolchlo: This may be nonstandard, but there are geometries in Antarctica that mess up our WebMercator projections that need to be pulled.

mojodna: Maybe better to start in web mercator to get all the area we can?:

```
val EARTH_RADIUS = 6378137
val clippedWorldExtent = Extent(-Math.PI * EARTH_RADIUS, -Math.PI * EARTH_RADIUS, Math.PI * EARTH_RADIUS, Math.PI * EARTH_RADIUS)
```

src/ingest/src/main/scala/osmesa/ingest/util/Caching.scala
```diff
@@ -0,0 +1,73 @@
...
+  private def s3exists(filename: String): Boolean =
+    s3client.doesObjectExist(bucket, prefix(filename) + "/_SUCCESS")
+
+  def orc(filename: String)(sparkjob: => DataFrame)(implicit ss: SparkSession): DataFrame = {
+    logger.warn("in cache func")
+    logger.warn("bucket: " ++ bucket)
+    logger.warn(prefix(filename))
+    logger.warn(fileUri(filename))
+    logger.warn("s3exists: " ++ s3exists(filename).toString)
+    val res = if (s3exists(filename)) {
```
mojodna: See https://github.com/azavea/osmesa/pull/47/files#diff-a2dba02322df99efde18635c62176f34 for some thoughts of preventing duplicate work.

src/ingest/src/main/scala/osmesa/ingest/Ingest.scala
mojodna: Ah, interesting. Can you add a note about that (ideally w/ a link)?

mojodna: (in the source, that is)

deployment/makefiles/scripts/configurations.json
jpolchlo: TBH, I'm not sure where in GTs configs we actually set the value.  It's nested down somewhere in the guts, or is provided by the typesafe config library.  We do set this value to improve performance, but it does lead to occasional spark task failures from cramming too much data into S3.  Not clear if lowering this value would clear the errors without lowering overall performance.

deployment/makefiles/Makefile
jpolchlo: I was seeing failures related to Netty when I was being stupid about my number of partitions, so I switched to the nio implementation.  The nio block transfer service was supposed to be a bit more performant, so I kept this config in, figuring it couldn't hurt.

deployment/makefiles/scripts/configurations.json
mojodna: Huh.

Longer-term, I'd like to investigate writing zips of vector tile pyramids to S3 (a preliminary implementation of this was in the Zeppelin notebook I sent over), since S3 writes can get expensive, both in terms of latency and $$$. I'll open an issue for discussing that.

deployment/makefiles/config-emr.mk
jpolchlo: These are in the configs, but you can see that I don't use them in the actual add-steps command in the Makefile.  I'm relying on the automatic allocation stuff now.  I can clean up these configs to clear up confusion later.

deployment/makefiles/scripts/configurations.json
jpolchlo: Noted.  (As I understand it, these are cluster-wide defaults.)

moradology:
superseded by #57

PR #58, opened 2018-04-18
Add a bounding box argument

moradology:
It is desirable to be able to produce vector tiles over some small subset of land without needing to produce  geographic extract according to that extent ahead of time. This PR adds the ability to provide a bounding box during `spark-submit` which filters out non-contained geometries.

### DIFF
--- a/src/ingest/src/main/scala/osmesa/ingest/Ingest.scala
+++ b/src/ingest/src/main/scala/osmesa/ingest/Ingest.scala
@@ -17,6 +17,7 @@ import org.apache.spark.rdd.RDD
 import org.apache.spark.sql._
 import org.apache.spark.sql.functions._
 import osmesa.ingest.util.Caching
+import org.apache.hadoop.fs._
 import vectorpipe._
 
 import scala.collection.mutable.ArrayBuffer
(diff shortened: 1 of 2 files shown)


src/ingest/src/main/scala/osmesa/ingest/Ingest.scala
```diff
@@ -90,18 +91,27 @@ object Ingest extends CommandApp(
...
     val changesetsO = Opts.option[String]("changesets", help = "Location of the ORC file containing changesets")
     val bucketO = Opts.option[String]("bucket", help = "S3 bucket to write VTs to")
     val prefixO = Opts.option[String]("key", help = "S3 directory (in bucket) to write to")
     val layerO = Opts.option[String]("layer", help = "Name of the output Layer")
     val maxzoomO = Opts.option[Int]("zoom", help = "Maximum zoom level for ingest (default=14)").withDefault(14)
-    val cacheDirO = Opts.option[String]("cache", help = "Location to cache ORC files").withDefault("")
     val pyramidF = Opts.flag("pyramid", help = "Pyramid this layer").orFalse
-    val filter = Opts.flag("filter", help = "GeoJSON multipolygon which contains all points for geoms to be processed")
+    val cacheDirO = Opts.option[String]("cache", help = "Location to cache ORC files").withDefault("")
+    val extentO = Opts.option[Extent]("extent", help = "Geographic extent for this layer").orNone
```
mojodna: What's the format of this option?

src/ingest/src/main/scala/osmesa/ingest/ProcessOSM.scala
```diff
@@ -44,33 +44,43 @@ object ProcessOSM {
...
-          (lead('timestamp, 1) over idByUpdated) as 'validUntil,
-          'uid,
-          'user,
-          'version,
-          'visible)
+    extent match {
+      case Some(e) =>
+        filteredHistory
+          .where('lat > e.ymin and 'lat < e.ymax)
+          .where('lon > e.xmin and 'lon < e.xmax)
```
mojodna: `>=`, etc?

src/ingest/src/main/scala/osmesa/ingest/Ingest.scala
moradology: Good call. Will add that to the documentation. This is the `fromString` I'm following: https://github.com/locationtech/geotrellis/blob/master/vector/src/main/scala/geotrellis/vector/Extent.scala#L34

PR #60, opened 2018-04-18
Add max age headers for s3 put requests

moradology:
Ideally, our tiles will be updated on some continuous basis. Long running client applications will need information about the expected duration of caching. This PR provides that information by way of a`max-age` setting on the `Cache-Control` header


moradology:
@mojodna as you can see, the max age is pretty specific to the non-zipped case and I've shoehorned the max-age logic into the `MakeTiles` function; I don't think it will cause a problem, but there's a faint code smell from the partial application I'm doing there...

moradology:
Stale; would be simpler to write fresh than fix this https://github.com/azavea/osmesa/issues/85

PR #84, opened 2018-08-15
Add documentation and examples for stats/vtile jobs

moradology:
Some of the steps involved in generating statistics and vector tiles are a bit murky. This PR aims to highlight the important bits


src/common/src/main/scala/osmesa/common/streaming/ChangeSource.scala
```diff
@@ -20,15 +20,16 @@ object ChangeSource extends Logging {
...
   def getSequence(baseURI: URI, sequence: Int): Seq[Element] = {
+    println(s"this URI: $baseURI this sequence: $sequence")
     val s = f"$sequence%09d".toArray
     val path =
       s"${s.slice(0, 3).mkString}/${s.slice(3, 6).mkString}/${s.slice(6, 9).mkString}.osc.gz"
 
-    logDebug(s"Fetching sequence $sequence")
-    val response = Http(baseURI.resolve(path).toString).asBytes
+    logInfo(s"Fetching sequence $sequence")
+    val response = Http(baseURI.resolve(path).toString).option(HttpOptions.allowUnsafeSSL).asBytes
```
mojodna: Is this a good idea? Under what conditions will we end up needing to access an HTTPS resource with a bad cert that isn't available over HTTP?

moradology: Given the fact that we've got a pretty strict contract about what data can look like, I doubt significant exploits are likely. I'm not super worried that this change stays, fwiw, so we can definitely remove it if it worries you.
As for why I added it: local `spark-submit` of streaming applications were failing for me because of cert problems which I wanted to work around

PR #87, opened 2018-08-23
ECS deployment for osmesa services

echeipesh:
- Provides resilience and monitoring
- Integrates with docker `HEALTHCHECK` command
- Publish/version osmesa service via ECR


src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetStreamProcessor.scala
```diff
@@ -183,7 +205,11 @@ object ChangesetStreamProcessor extends CommandApp(
               this.partitionId = partitionId
               this.version = version
 
-              connection = DriverManager.getConnection(s"jdbc:${databaseUri.toString}")
+              connection = (databaseUser, databasePass).mapN { (usr, pass) =>
+                DriverManager.getConnection(s"jdbc:${databaseUri.toString}", usr, pass)
```
mojodna: Username and password can be included in a JDBC URI as `?user=<username>&password=<password>`, which can simplify this a bit.

Ideally we'd take username and password from the auth part of a URI and copy it in there; that makes it easy to extract from an environment variable: `... --database-uri $DATABASE_URL`

deployment/sql/checkpoints.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE checkpoints (
```
mojodna: https://github.com/azavea/osmesa/blob/master/src/analytics/Makefile should also be moved into `deployment/sql` and an entry added for `checkpoints` in the `db/all` target.

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetStreamProcessor.scala
echeipesh: I see that the URI currently used is `postgresql://[host]/[db]?user=[username]&password=[password]` which doesn't square with `postgresql://[user[:password]@][netloc][:port][/dbname][?param1=value1&...]` format mentioned https://www.postgresql.org/docs/9.3/static/libpq-connect.html

We can/should iterate through these, but wondering why the diff.

mojodna: You and me both. It looks like a JDBC wart where auth is ignored when passed in the typical part of a URI.

PR #92, opened 2018-09-18
Parse database URI for username and password

echeipesh:
Response to comments in: https://github.com/azavea/osmesa/pull/87

- [x] Parse URI for user/password
- [x] Use `DATABASE_URL` env variable as fallback
- [x] streaming README
- [x] Test deployment live


src/analytics/src/main/scala/osmesa/analytics/oneoffs/AugmentedDiffStreamProcessor.scala
```diff
@@ -272,13 +254,7 @@ object AugmentedDiffStreamProcessor extends CommandApp(
...
             this.partitionId = partitionId
             this.version = version
-
-            connection = (databaseUser, databasePass).mapN { (usr, pass) =>
-              DriverManager.getConnection(s"jdbc:${databaseUri.toString}", usr, pass)
-            }.getOrElse {
-              DriverManager.getConnection(s"jdbc:${databaseUri.toString}")
-            }
-
+            connection = DBUtils.getJdbcConnection(databaseUri)
```
echeipesh: So tempting to make this method extension on `URI` but those tend to become magical so I'm resisting the urge.

echeipesh: ex: `databaseUri.getJdbcConnection()`

deployment/streaming/Makefile
```diff
@@ -1,46 +1,115 @@
...
+ECR_REPO := ???
+AWS_LOG_GROUP := osm-stats-stream
+AWS_REGION := us-east-1
+
+# Shared properties
+AUGDIFF_START := 3164313
+AUGDIFF_SOURCE := s3://nome-osmesa/augmented-diffs/
+AUGDIFF_LOCAL_SOURCE := s3://mojodna-temp/augdiffs/
+CHANGESET_START := 168249
+CHANGESET_SOURCE := http://10.0.1.244/replication/replication/changesets/
```
mojodna: Let's remove custom URIs in favor of keeping them in local configs.

deployment/streaming/docker-compose.deploy.yml.tpl
```diff
@@ -0,0 +1,44 @@
...
+        awslogs-stream-prefix: augdiff
+  changeset-stream:
+    image: ${ECR_REPO}:latest
+    command: >
+      /spark/bin/spark-submit --driver-memory 4096m --class osmesa.analytics.oneoffs.ChangesetStreamProcessor /opt/osmesa-analytics.jar
+      --changeset-source ${CHANGESET_SOURCE}
+      --start-sequence ${CHANGESET_START}
+      --database-uri ${DB_URI}
+      --database-user ${DB_USER}
+      --database-pass ${DB_PASS}
```
mojodna: User / password are no longer necessary as separate args, right? (Same for above.)

deployment/streaming/docker-compose.deploy.yml.tpl
```diff
@@ -0,0 +1,44 @@
+version: '3.0'
+services:
+  augdiff-stream:
+    image: ${ECR_REPO}:latest
+    command: >
+      /spark/bin/spark-submit --driver-memory 4096m --class osmesa.analytics.oneoffs.AugmentedDiffStreamProcessor /opt/osmesa-analytics.jar
```
mojodna: Does it make sense to submit these to a shared local master + worker (w/ reservations on partial capacity) rather than running 2 in "local" mode? (For logging / better capacity management??)

PR #112, opened 2018-11-21
Add coastline stats to analytics package

jpolchlo:
We have tracked roads and waterways.  This PR aims to include coastline features on a similar footing.


jpolchlo:
@mojodna: I suspect that the sql in `deployments` should be updated to include the coastline stats as well?

src/common/src/main/scala/osmesa/common/functions/osm/package.scala
```diff
@@ -224,6 +224,10 @@ package object osm {
     (_: Map[String, String]).contains("highway")
   }
 
+  val isCoastline: UserDefinedFunction = udf {
+    (_: Map[String, String]).getOrElse("natural", "no").toLowerCase == "coastline"
```
mojodna: Seriously tiny nit: I would default to `""` rather than `"no"` here. It doesn't change anything, but reading through, it made me wonder, "why would it be 'no'?"

src/analytics/src/main/scala/osmesa/analytics/stats/CalculateStats.scala
```diff
@@ -349,7 +349,7 @@ object CalculateStats {
           (wayRelations ++ ways).
             flatten.
             foldLeft(false) { case (acc, (_, upstreamInfos)) =>
-              acc || upstreamInfos.hasTopic(StatTopics.ROAD) || upstreamInfos.hasTopic(StatTopics.WATERWAY)
```
mojodna: This file is no longer relevant.

src/analytics/src/main/scala/osmesa/analytics/stats/StatCounter.scala
```diff
@@ -67,13 +67,17 @@ class StatCounter private (
   def buildingsModified: Int = countModified(StatTopics.BUILDING)
```
mojodna: Also not relevant

src/analytics/src/main/scala/osmesa/analytics/stats/StatTopics.scala
```diff
@@ -5,14 +5,15 @@ import osmesa.analytics.Constants
 import scala.collection.mutable
```
mojodna: Also not relevant

PR #114, opened 2018-12-06
`constructGeometries` should produce a geom column with type `Geometry`

jpolchlo:
Due to the order of joining in `ProcessOSM.constructGeometries`, the resulting frame has a `geom` column of type `point`.  Since JTS frames only provides functions to make geometries more specific and reversing the join order is not allowed (for some reason), I've provided a local udf to do the conversion.


mojodna:
A UDF matching GeoMesa's conventions was added as part of https://geomesa.atlassian.net/browse/GEOMESA-2454 (same/similar issue).

PR #157, opened 2019-08-01
Staging deployment and Railway statistics

jpolchlo:
We've been somewhat primitive in the way that we deploy and test different versions of OSMesa.  A request arrived where we finally could not ignore the fact that a staging environment was needed, as it showed our terrible lack of sophistication in the way that we were storing and dealing with statistics.  Previously, every small change to the stats would require a change in table schema, which required a complex backfill process into an existing and actively used DB table.  Following which, we would need to restart a streaming update process and then infill.  All because we were amending a live table that was feeding a process visible to end users.

This PR will split our deployments into staging and production environments.  This considerably simplifies changes, as they can be private to a development team.  Rolling out a new stat will then be accomplished by deleting the contents of any existing tables, repopulating the tables, restarting a streaming process (which can take as long as needed to catch up), and then redeploying dependent services.

This split was prompted by the need for railway statistics reporting which are also added to this PR.

The changes here are mirrored by those in azavea/osmesa-stat-server#22.


src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetORCCreator.scala
```diff
@@ -0,0 +1,175 @@
+package osmesa.analytics.oneoffs
```
jpolchlo: @mojodna This is a separate utility for updating our changeset ORC files from the changeset stream.  This automatically finds the sequence number to start changeset stream from.  It seems like this is an improvement from the current state of affairs, but reasonable people can disagree.  You have any thoughts on this?

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetORCCreator.scala
```diff
@@ -0,0 +1,175 @@
...
+ * spark-submit \
+ *   --class osmesa.analytics.oneoffs.ChangesetORCUpdater \
+ *   ingest/target/scala-2.11/osmesa-analytics.jar \
+ *   --changeset-source http://somewhere/diffs/ \
+ *   --database-url $DATABASE_URL
+ */
+object ChangesetORCUpdater
+  extends CommandApp(
+    name = "osmesa-changeset-stats-updater",
+    header = "Update statistics from augmented diffs",
```
mojodna: Copy pasta.

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetORCCreator.scala
mojodna: This looks very useful! So much so that maybe it belongs in VP with the other core OSM-related functionality?

I think the name could be clearer. Something along the lines of "merge subsequent changesets".

I also wonder if it makes sense to make the ORC argument optional and allow either sequence numbers (if one has those handy for some reason) or timestamps instead. It would also be nice to include the start/end sequences in the output filename (since there's no easily-accessible metadata otherwise), but in a way that can be overridden.

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetORCCreator.scala
```diff
@@ -0,0 +1,175 @@
...
+          .repartition(1)
+          .write
+          .orc(outputURI.toString)
+
+        spark.stop()
+      }
+    }
+)
+
+object ChangesetORCUpdaterUtils {
```
mojodna: This is some nice common functionality for changesets; belongs in VP?

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetORCCreator.scala
```diff
@@ -0,0 +1,175 @@
...
+
+  // implicit val TimestampFormat : Decoder[Timestamp] = new Decoder[Timestamp] {
+  //   private val sdf = new java.text.SimpleDateFormat("yyyy-MM-dd HH:mm:ss XXX")
+
+  //   override def apply(c: HCursor): Decoder.Result[Timestamp] =
+  //     c.get[String]("last_run").map{ str =>
+  //       val splt = str.split(".")
+  //       Timestamp.from(sdf.parse(splt(0) ++ splt(1).split(" ")(1)).toInstant)
+  //     }
+  // }
```
mojodna: Is this still useful?

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetORCCreator.scala
```diff
@@ -0,0 +1,175 @@
...
+      Http(baseURI.resolve(path).toString).asString
+
+    val state = yaml.parser
+      .parse(response.body)
+      .leftMap(err => err: Error)
+      .flatMap(_.as[Sequence])
+      .valueOr(throw _)
+
+    state
+  }
```
mojodna: Just return from `yaml.parser`?

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetORCCreator.scala
```diff
@@ -0,0 +1,175 @@
...
+
+  def estimateSequenceNumber(modifiedTime: Instant, baseURI: URI): Long = {
+    val current = getCurrentSequence(baseURI)
+    val diffMinutes = (current.last_run.toInstant.getMillis/1000 - modifiedTime.getEpochSecond) / 60
+    current.sequence - diffMinutes
+  }
+
+  def findSequenceFor(modifiedTime: Instant, baseURI: URI): Long = {
+    var guess = estimateSequenceNumber(modifiedTime, baseURI)
+    val target = org.joda.time.Instant.parse(modifiedTime.toString)
```
mojodna: Is Joda-Time necessary here, or does `java.time` have what we need now?

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetORCCreator.scala
```diff
@@ -0,0 +1,175 @@
...
+
+    while (getSequence(baseURI, guess).last_run.isAfter(target)) { guess -= 1 }
+    while (getSequence(baseURI, guess).last_run.isBefore(target)) { guess += 1 }
+
+    getSequence(baseURI, guess).sequence
```
mojodna: This is to confirm that the sequence exists?

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetStatsCreator.scala
```diff
@@ -111,17 +111,17 @@ object ChangesetStatsCreator
                        'tags.getItem("created_by") as 'editor,
                        'uid,
                        'user,
-                       'created_at,
+                       'createdAt,
```
mojodna: Capitalization depends on whether an `osm2orc` changeset file or a replication-sourced DF is the source here (sorry, that's my bad).

`osm2orc` uses underscores for parity with the changeset XML. Perhaps `ChangesetSource` should follow suit? Or we should have a function that normalizes column names to use before queries like this one.

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetORCCreator.scala
jpolchlo: Joda time was required to deal with the fractional seconds portion of the date/time strings provided by the state files from overpass.  Java time can deal with milliseconds, which are not the same as fractional seconds.  Or so it seems.  Perhaps testing is required to see if we need to drop this part of the date/time string?

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetStatsCreator.scala
jpolchlo: Let's normalize the column names.

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetORCCreator.scala
jpolchlo: I thought so, but worth paring down to the bare bones.

jpolchlo:
Managed a successful ingest of the basic stats.  These are currently populating the staging DB.  Will take the time to see that we have identified any railway stats, but seems like the JSON stuff works, after slight modification (see last commit).  I'll try and fire up the streaming job in the morning.

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetORCCreator.scala
jpolchlo: This comment made less sense to me before I realized that I forgot to union the existing changeset ORC to the streamed input.  That's fixed.  Do you still hath concerns about this?

I think what you really want by this comment will mostly be satisfied by bringing the `ChangesetORCUpdaterUtils` into VP's `ChangesetSource`?  Is that somewhat accurate?

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetORCCreator.scala
```diff
@@ -0,0 +1,181 @@
...
+
+        val options = Map(
+          Source.BaseURI -> changesetSource.toString,
+          Source.StartSequence -> startSequence.toString,
+          Source.EndSequence -> (endSequence + 1).toString // sequence range is (]; end sequence is exclusive
+        )
+
+        val changesets = spark.read.format(Source.Changesets).options(options).load
+        changesets
+          .drop("comments", "sequence")
```
jpolchlo: I wonder if it's better to drop these columns, or to somehow preserve them, though, they aren't currently in the table schema...

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetORCCreator.scala
jpolchlo: This process is to ensure that we return the first changeset sequence number that comes after the date provided.  That should allow the streaming operation to pick up without missing or duplicating anything.

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetStatsCreator.scala
jpolchlo: I'm just going to assume that we're using the osm2orc naming structure (partly because of `model.Changeset`).  If that is going to get us in trouble elsewhere, speak now or forever holster your piece.

src/analytics/src/main/scala/osmesa/analytics/stats/ChangesetStatsForeachWriter.scala
```diff
@@ -41,7 +41,7 @@ class ChangesetStatsForeachWriter(databaseUri: URI,
       |  measurements = (
       |    SELECT jsonb_object_agg(key, value)
       |    FROM (
-      |      SELECT key, sum(value::numeric) AS value
+      |      SELECT key, sum((value->>0)::numeric) AS value
```
jpolchlo: This was aggravating to figure out.  Shame SQL doesn't have static type analysis.

jpolchlo:
Can positively report that railway stats are in the table.  Also, we are now automatically adding checkpoints to the DB so that we can more smoothly start up the streaming update process

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetMetadataUpdater.scala
```diff
@@ -80,7 +80,8 @@ object ChangesetMetadataUpdater
               import ss.implicits._
 
               val options = Map(
-                Source.BaseURI -> changesetSource.toString
+                Source.BaseURI -> changesetSource.toString,
+                Source.ProcessName -> "ChangesetStream"
```
mojodna: This doesn't achieve what you were hoping it did. `Source.ProcessName` is only used by _streaming_ sources (in `commit` after processing batches); this is a regular source.

Instead, we should have something that can be executed after `forEachPartition` completes to update the checkpoints appropriately.

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetORCCreator.scala
```diff
@@ -0,0 +1,204 @@
...
+    val target = org.joda.time.Instant.parse(modifiedTime.toString)
+
+    while (getSequence(baseURI, guess).last_run.isAfter(target)) { guess -= 1 }
+    while (getSequence(baseURI, guess).last_run.isBefore(target)) { guess += 1 }
+
+    getSequence(baseURI, guess).sequence
+  }
+
+  def saveLocations(procName: String, sequence: Int, databaseURI: URI) = {
+    var connection = null.asInstanceOf[Connection]
```
mojodna:
```suggestion
    var connection: Connection = _
```

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetORCCreator.scala
mojodna: Largely, yes, but I'm also quibbling about the name of this command. May I propose `MergeChangesets` instead of `ChangesetORCCreator`?

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetORCCreator.scala
mojodna: I would love to preserve them in the combined output. We don't do anything with comments (and they're not in the dump we typically convert), but I can see eventual value. For `sequence`, we can let it be `null` for rows that came from a dump.

In theory, the probably `sequence` could be calculated based on the timestamp, but probably best to leave that to downstream consumers.

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetORCCreator.scala
```diff
@@ -0,0 +1,204 @@
...
+  }
+
+  private val formatter = DateTimeFormat.forPattern("y-M-d H:m:s.SSSSSSSSS Z")
+
+  private implicit val dateTimeDecoder: Decoder[DateTime] =
+    Decoder.instance(a => a.as[String].map(DateTime.parse(_, formatter)))
+
+  case class Sequence(last_run: DateTime, sequence: Long)
+
+  def getCurrentSequence(baseURI: URI): Sequence = {
```
mojodna: I like this signature more than VP's current `def getCurrentSequence(baseURI: URI): Option[Int]`. Returning `Option[Sequence]` instead of throwing if it can't be determined would be nice though.

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetORCCreator.scala
```diff
@@ -0,0 +1,204 @@
...
+    val state = yaml.parser
+      .parse(response.body)
+      .leftMap(err => err: Error)
+      .flatMap(_.as[Sequence])
+      .valueOr(throw _)
+
+    state
+  }
+
+  def getSequence(baseURI: URI, sequence: Long): Sequence = {
```
mojodna: 👍 

VP's `getSequence` should be renamed to `getChangesetsForSequence`, eh?

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetORCCreator.scala
mojodna: 👍

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetORCCreator.scala
mojodna: Let's just keep using Joda-Time to avoid manipulating the string..  It seems like there should be a way to use `java.time` (with a `DateTimeFormatter`?) but it's certainly not straightforward...

```
scala> import java.time.Instant
import java.time.Instant

scala> Instant.parse("2019-08-06 22:11:02.531677000 +00:00")
java.time.format.DateTimeParseException: Text '2019-08-06 22:11:02.531677000 +00:00' could not be parsed at index 10
  at java.base/java.time.format.DateTimeFormatter.parseResolved0(DateTimeFormatter.java:2046)
  at java.base/java.time.format.DateTimeFormatter.parse(DateTimeFormatter.java:1948)
  at java.base/java.time.Instant.parse(Instant.java:395)
  ... 28 elided
```

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetStatsCreator.scala
```diff
@@ -27,6 +28,16 @@ object ChangesetStatsCreator
...
 
+      val changesetBaseOpt =
+        Opts
+          .option[URI](
+          "changeset-stream",
+          short = "c",
+          metavar = "uri",
+          help = "HTTP Location of replication changesets"
+        )
+        .validate("Changeset source must have trailing '/'") { _.getPath.endsWith("/") }
```
mojodna: Can we drop this requirement and update URIs to match within `ChangesetSource`?

This can also have a default value of `https://planet.openstreetmap.org/replication/changesets/` (which should really be a default value in `ChangesetSource`, as it gets duplicated all over right now.

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetStatsCreator.scala
```diff
@@ -37,12 +48,20 @@ object ChangesetStatsCreator
...
             import spark.implicits._
 
+            val logger = org.apache.log4j.Logger.getLogger(getClass())
+
             val history = spark.read.orc(historySource)
+
+            val augdiffEndSequence = {
+              val t = history.select(max('timestamp)).first.getAs[java.sql.Timestamp](0).toInstant
+              ((t.getEpochSecond - 1347432900) / 60).toInt
+            }
```
mojodna: Move into `AugmentedDiffSource`?

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetStatsCreator.scala
mojodna: osm2orc doesn't match `model.Changeset`.  That hadn't really been a problem before since we never actually cast a `DataFrame` of changesets to `Dataset[Changeset]` and the stream emits `Dataset[Row]` (`DataFrame`) even though it uses `Changeset` internally...

This suggests to me that the normalization function should go underscores -> camel case. Though the counter-argument is that certain output formats (JDBC, for one) lowercase column names before writing.

mojodna: Ugh.

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetStatsCreator.scala
```diff
@@ -155,6 +178,16 @@ object ChangesetStatsCreator
...
 
+            // Distributing these writes to the executors to avoid no suitable driver errors on master node
+            logger.warn(s"Writing AugmentedDiffStream sequence number as $augdiffEndSequence to $databaseUrl")
+            spark.sparkContext.parallelize(Seq(databaseUrl)).foreach { uri =>
+              ChangesetORCUpdaterUtils.saveLocations("AugmentedDiffStream", augdiffEndSequence, uri)
+            }
+            logger.warn(s"Writing ChangesetStream sequence number as $changesetsEndSequence to $databaseUrl")
+            spark.sparkContext.parallelize(Seq(databaseUrl)).foreach { uri =>
+              ChangesetORCUpdaterUtils.saveLocations("ChangesetStream", changesetsEndSequence, uri)
+            }
```
mojodna: Ah, these are the calls I was I was aspiring to in https://github.com/azavea/osmesa/pull/157/files#r311292846

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetStatsCreator.scala
```diff
@@ -155,6 +178,16 @@ object ChangesetStatsCreator
                 }
               })
 
+            // Distributing these writes to the executors to avoid no suitable driver errors on master node
```
mojodna: _shakes fist at JDBC!!!_

Does invoking the name of the driver help? This is a weird workaround to need to remember to do.

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetStatsUpdater.scala
```diff
@@ -91,7 +91,8 @@ object ChangesetStatsUpdater
             import ss.implicits._
 
             val options = Map(
-              Source.BaseURI -> augmentedDiffSource.toString
+              Source.BaseURI -> augmentedDiffSource.toString,
+              Source.ProcessName -> "AugmentedDiffStream"
```
mojodna: See https://github.com/azavea/osmesa/pull/157/files#r311292846

src/analytics/src/main/scala/osmesa/analytics/stats/ChangesetStatsForeachWriter.scala
mojodna: Did this not previously work? Values should be scalar.

My objection to `(value->>0)::numeric` is that it casts a JSON number to a string (`->>0` vs. `->0`) and then back to a number vs. just casting the JSON number to a numeric.

```
seth=# select pg_typeof(('[12.4]'::jsonb)->0);
 pg_typeof
═══════════
 jsonb
(1 row)
seth=# select pg_typeof(('[12.4]'::jsonb)->>0);
 pg_typeof
═══════════
 text
(1 row)
```

src/analytics/src/main/scala/osmesa/analytics/stats/ChangesetStatsForeachWriter.scala
```diff
@@ -54,7 +54,7 @@ class ChangesetStatsForeachWriter(databaseUri: URI,
       |  counts = (
       |    SELECT jsonb_object_agg(key, value)
       |    FROM (
-      |      SELECT key, sum(value::numeric) AS value
+      |      SELECT key, sum((value->>0)::numeric) AS value
```
mojodna: Samesies.

src/analytics/src/main/scala/osmesa/analytics/stats/ChangesetStatsForeachWriter.scala
mojodna: Can you dump `measurements` from one of the rows? I want to understand how values became arrays.

mojodna: N/m. I was developing against pg 11; prod is 10.6.

10.6:

```
=> select key, sum(value::numeric) from jsonb_each((select counts from changesets where id = <something>)) group by key;
ERROR:  cannot cast type jsonb to numeric
LINE 1: select key, sum(value::numeric) from jsonb_each((select coun...
```

11.4:

```
=# select key, sum(value::numeric) from jsonb_each((select counts from changesets where id = 3029981)) group by key;
      key       │ sum
════════════════╪═════
 roads_added    │   2
 roads_modified │  18
(2 rows)
```

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetStatsCreator.scala
jpolchlo: I agree that it's weird.  But it appears to be a classpath thing.  Works fine locally, where everything is installed in one place.  Something about the setup on YARN is making this necessary.  Found no pointers of any use on the world wide web.

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetMetadataUpdater.scala
jpolchlo: It seems like this (and the mirrored comment below) can just be omitted?  I've been really confused while trying to figure out how the various classes know where to look for their current sequence number.  I'm also not clear why this isn't a streaming source.  Do we ever `spark.readStream` anywhere?  The nature of the sources supplied by VP is a continuing source of confusion that I would love to clear up.

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetStatsCreator.scala
mojodna: _le sigh_

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetMetadataUpdater.scala
mojodna: `StreamingChangesetMetaUpdater` uses `spark.readStream`. `readStream` creates a streaming source (using the "MicroBatchReader" implementation, producing an unending stream of batched updates), `read` creates a non-streaming source (producing a single `DataFrame` with a finite range of updates).

Start sequence calculation / recovery is handled in 2 places (which could probably be merged, causing `Source.ProcessName` to be used in non-streaming contexts, although only streaming sources will **update** the checkpoint):

* [`ReplicationStreamMicroBatchReader`](https://github.com/geotrellis/vectorpipe/blob/cbeb773ed9dc6bf211221611f136047c435451e4/src/main/scala/vectorpipe/sources/ReplicationStreamMicroBatchReader.scala#L42-L49) -- when `readStream` is used. Resolution order is: 1) `checkpoints` table, 2) `Source.StartSequence` option
* [`ReplicationReader`](https://github.com/geotrellis/vectorpipe/blob/cbeb773ed9dc6bf211221611f136047c435451e4/src/main/scala/vectorpipe/sources/ReplicationReader.scala#L34-L40) -- when `read` is used. Resolution order is: 1) `Source.StartSequence` option, 2) `getCurrentSequence`

For the `ReplicationStreamMicroBatchReader`, the 3rd resolver **should** be `getCurrentSequence`.

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetStatsCreator.scala
jpolchlo: Noted in geotrellis/vectorpipe#111.

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetStatsCreator.scala
jpolchlo: Is this relevant to OSM?  This may need to be removed, now that you mention it...

mojodna: It is. Overpass replication sequences are the same across all instances, regardless of source.

jpolchlo:
This now depends on geotrellis/vectorpipe#109

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetStatsCreator.scala
jpolchlo: Found in https://github.com/geotrellis/vectorpipe/pull/109/commits/d4c5d36430be79cb438dcc38954bb28adc9a7901

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetORCCreator.scala
jpolchlo: How about this: https://github.com/geotrellis/vectorpipe/pull/109/files#diff-73c6a8d6784213e9174e19e1215482c8R31 ?

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetORCCreator.scala
jpolchlo: Note https://github.com/geotrellis/vectorpipe/blob/d4c5d36430be79cb438dcc38954bb28adc9a7901/src/main/scala/vectorpipe/sources/ChangesetSource.scala lines 73 and 95

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetORCCreator.scala
jpolchlo: The proposed change generates a `local variables must be initialized` error.  But replace the underscore with `???` and it compiles.  Not sure if it runs, but it does compile.

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetORCCreator.scala
jpolchlo: This item isn't demanding action at present, so I'll ignore it for the purposes of this PR.  But if there's something specific to do in relation to this point, make an issue in VP?

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetORCCreator.scala
jpolchlo: Maybe most readable option is `var connection: Connection = null`?

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetORCCreator.scala
jpolchlo: I did opt for MergeChangesets, FYI.

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetStatsCreator.scala
jpolchlo: As far as the current PR goes, this isn't exactly needed.  For the moment, it seems as if wee could provide, in VP, a method (or two) in `ChangesetSource` that could be used to ensure that the names are all either in camel case or in snake case.  For now, we do this conversion manually ([here](https://github.com/azavea/osmesa/blob/3349bf5c6c3c46c2bebcb86342e176ba1dde7020/src/analytics/src/main/scala/osmesa/analytics/oneoffs/MergeChangesets.scala#L100-L113)), but automatic means for converting should be provided.  I've created an issue [here](https://github.com/geotrellis/vectorpipe/issues/113) to capture this need.

If that's sufficient, then I suggest we just move on and not worry too much about the concern in this comment.  I'll leave a note in the code to clean this up?

deployment/streaming/scripts/batch-generate-edit-histograms.sh
```diff
@@ -0,0 +1,59 @@
...
+	      {
+	        "Args": [
+	          "spark-submit",
+	          "--deploy-mode", "cluster",
+	          "--class", "osmesa.analytics.oneoffs.FacetedEditHistogramTileCreator",
+	          "--conf", "spark.executor.memoryOverhead=2g",
+	          "--conf", "spark.sql.shuffle.partitions=2000",
+	          "--conf", "spark.speculation=true",
+	          "${OSMESA_ANALYTICS_JAR}",
+	          "--history", "${NOME_HISTORY_ORC}",
```
mojodna: `${HISTORY_ORC}`

deployment/streaming/scripts/batch-generate-footprints.sh
```diff
@@ -0,0 +1,61 @@
...
+	      {
+	        "Args": [
+	          "spark-submit",
+	          "--deploy-mode", "cluster",
+	          "--class", "osmesa.analytics.oneoffs.FootprintCommand",
+	          "--conf", "spark.executor.memoryOverhead=2g",
+	          "--conf", "spark.sql.shuffle.partitions=2000",
+	          "--conf", "spark.speculation=true",
+	          "${OSMESA_ANALYTICS_JAR}",
+	          "--history", "${NOME_HISTORY_ORC}",
```
mojodna: ☝️

deployment/streaming/scripts/batch-generate-footprints.sh
```diff
@@ -0,0 +1,61 @@
...
+	          "--changesets", "${NOME_CHANGESETS_ORC",
```
mojodna: This is also missing a closing `}`

deployment/streaming/scripts/configurations.json
```diff
@@ -0,0 +1,88 @@
+[
```
mojodna: Is this needed?

jpolchlo: Used [here](https://github.com/azavea/osmesa/blob/baca909e376116350fbb0cf60e32889a9194f0b3/deployment/streaming/scripts/batch-process.sh#L15).  Perhaps it's not needed?  I may have just included this as a test.  I don't entirely recall.

src/analytics/src/main/scala/osmesa/analytics/oneoffs/MergeChangesets.scala
```diff
@@ -0,0 +1,163 @@
...
+import java.time.Instant
+import scalaj.http.Http
+
+/*
+ * Usage example:
+ *
+ * sbt "project analytics" assembly
+ *
+ * spark-submit \
+ *   --class osmesa.analytics.oneoffs.ChangesetORCUpdater \
```
mojodna: This should match the object name.

src/analytics/src/main/scala/osmesa/analytics/oneoffs/MergeChangesets.scala
```diff
@@ -0,0 +1,163 @@
...
+ *   ingest/target/scala-2.11/osmesa-analytics.jar \
+ *   --changesets http://location/of/changeset/replications \
+ *   --end-time 1970-01-01T13:00:00Z
+ *   s3://path/to/history.orc
+ *   s3://path/to/output.orc
+ */
+object MergeChangesets
+  extends CommandApp(
+    name = "osmesa-changeset-orc-updater",
```
mojodna: This should also match-ish the object name.

src/analytics/src/main/scala/osmesa/analytics/oneoffs/MergeChangesets.scala
```diff
@@ -0,0 +1,163 @@
...
+          .repartition(1)
+          .write
+          .orc(outputURI.toString)
+
+        spark.stop()
+      }
+    }
+)
+
+object MergeChangesetUtils {
```
mojodna: Since this is used elsewhere, should it be outside of the `oneoffs` package?

src/analytics/src/main/scala/osmesa/analytics/oneoffs/ChangesetMetadataUpdater.scala
mojodna: I'll throw up a PR to fix this in VP.

deployment/streaming/scripts/configurations.json
mojodna: When I've run things in the past, I haven't needed this, fwiw. It looks like it may be GT boilerplate for EMR.

src/analytics/src/main/scala/osmesa/analytics/oneoffs/MergeChangesets.scala
jpolchlo: There is so much reorganization of this repo that needs to happen.  I can see breaking out oneoffs into streaming, batch, and utilities.

deployment/streaming/scripts/configurations.json
jpolchlo: It's probably not needed.  I was throwing whatever at the wall to get things to run to completion, so I think I can drop it now that things are all sussed out.

jpolchlo:
Seems like the long chain of discussion here has finally played out.  Merging.

PR #160, opened 2019-09-25
Streaming AOI monitor service

jpolchlo:
We are developing a facility to track areas of interest and provide periodic updates to subscribed users as to edits occurring in those regions.  This PR provides the component which will monitor and aggregate those changes and queue up the information that will be conveyed to the users.

We intend to deploy this service via ECS scheduled tasks at preset intervals (monthly, weekly, daily), and each interval may aggregate and report its results differently.  Email notifications will be handled by a different service.

### DIFF
--- a/src/analytics/src/main/scala/osmesa/analytics/oneoffs/StreamingAOIMonitor.scala
+++ b/src/analytics/src/main/scala/osmesa/analytics/oneoffs/StreamingAOIMonitor.scala
@@ -4,13 +4,17 @@ import cats.data.{Validated, ValidatedNel}
 import cats.implicits._
 import com.monovore.decline.{CommandApp, Opts}
 import geotrellis.vector._
-import org.apache.spark.sql.SparkSession
-import org.apache.spark.sql.functions.{explode, from_unixtime, to_timestamp, udf}
+import org.apache.spark.internal.Logging
+import org.apache.spark.sql.{Dataset, SparkSession}
+import org.apache.spark.sql.functions.{collect_list, collect_set, count, explode, size, udf}
 import org.locationtech.geomesa.spark.jts.{udf => _, _}
 import org.locationtech.jts.{geom => jts}
 import org.locationtech.jts.geom.prep._
 import osmesa.analytics.Analytics
-import vectorpipe.sources.Source
+import osmesa.analytics.stats._
+import osmesa.analytics.stats.functions._
+import vectorpipe.functions._
+import vectorpipe.sources.{AugmentedDiffSource, Source}
 import vectorpipe.util.DBUtils
 
 import java.net.URI
@@ -28,7 +32,7 @@ object StreamingAOIMonitor extends CommandApp(
     val intervalOpt =
       Opts
         .option[String]("interval", help = "Period of time to aggregate over (d=daily, w=weekly)")
-        .validate("Must be one of 'd', 'w'") { arg: String => arg.length == 1 && "dwm".contains(arg.apply(0)) }
+        .validate("Must be one of 'd', 'w'") { arg: String => arg.length == 1 && "dw".contains(arg.apply(0)) }
 
     val databaseUrlOpt =
       Opts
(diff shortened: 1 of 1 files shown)


src/analytics/src/main/scala/osmesa/analytics/oneoffs/StreamingAOIMonitor.scala
```diff
@@ -0,0 +1,176 @@
...
+    val augmentedDiffSourceOpt =
+      Opts
+        .option[URI](
+        "augmented-diff-source",
+        short = "a",
+        metavar = "uri",
+        help = "Location of augmented diffs to process"
+      )
+
+    val startSequenceOpt =
```
CloudNiner: In the interest of Ya Aint Gonna Need It, the specs for our initial MVP are that the start / end sequences necessary are generated from the interval parameter passed by the scheduled task. Given that, we should probably remove these two sequence params and just generate the start and end sequence values from the interval, using `now` as the end.

src/analytics/src/main/scala/osmesa/analytics/oneoffs/StreamingAOIMonitor.scala
```diff
@@ -0,0 +1,176 @@
...
+import geotrellis.vector.io.json._
+import spray.json._
+
+object StreamingAOIMonitor extends CommandApp(
+  name = "streaming-aoi-monitor",
+  header = "Streaming AOI Monitor",
+  main = {
+    val intervalOpt =
+      Opts
+        .option[String]("interval", help = "Period of time to aggregate over (d=daily, w=weekly, m=monthly)")
```
CloudNiner: There was a conversation in the doc a day or so ago with Ross about removing monthly for now in order to simplify the initial ask. Let's do that here so that the extra option doesn't propagate further into the work we're doing.

src/analytics/src/main/scala/osmesa/analytics/oneoffs/StreamingAOIMonitor.scala
```diff
@@ -0,0 +1,176 @@
...
+      endSequence
+        .map(s => Map(Source.EndSequence -> s.toString))
+        .getOrElse(Map.empty[String, String])
+
+      val diffs = spark
+        .readStream
+        .format(Source.AugmentedDiffs)
+        .options(options)
+        .load
+        .withColumn("watermark", to_timestamp(from_unixtime('sequence * 60 + 1347432900)))
```
CloudNiner: What are the values of the constants here? It would be helpful if these were named to more easily understand the code's intent.

src/analytics/src/main/scala/osmesa/analytics/oneoffs/StreamingAOIMonitor.scala
```diff
@@ -0,0 +1,176 @@
...
+        .withWatermark("watermark", "0 seconds")
+
+      val aoiTag = udf { g: jts.Geometry => aoiIndex(g).toList }
+
+      val grouped = diffs.withColumn("aoi", explode(aoiTag('geom))).groupBy('aoi)
+
+      // DO SOMETHING WITH THE DATA.
+      // Aggregations will likely differ with
```
CloudNiner: Incomplete comment?

src/analytics/src/main/scala/osmesa/analytics/oneoffs/StreamingAOIMonitor.scala
jpolchlo: For testing purposes, it makes the most sense to leave these in for the time being.  I agree that we will ultimately draw the sequence numbers from different sources, but for now, being able to pull a known set of diffs is useful.

CloudNiner: Works for me for now. Given that though, could we add the logic that converts an interval to start/end sequence values now? If interval is provided, it would override any values passed to startSequence and endSequence or vice versa.

src/analytics/src/main/scala/osmesa/analytics/oneoffs/StreamingAOIMonitor.scala
jpolchlo: I left the iron on.  Had to leave in a hurry.

src/analytics/src/main/scala/osmesa/analytics/oneoffs/StreamingAOIMonitor.scala
CloudNiner: 👍

src/analytics/src/main/scala/osmesa/analytics/oneoffs/StreamingAOIMonitor.scala
jpolchlo: The way I see it, the interval doesn't only affect the selection of start/stop sequences, but also how the data are aggregated.  (That was what that abortive comment was going to get at before I saw something shiny and forgot to finish typing.)  I want to have the control during testing to pick a window of time to look at, and also the freedom to select how it will be presented.  Or so the thinking goes.

CloudNiner: Ok. It makes sense to keep the start/stop params in for now, but I'm not sure what that means for the comment above though in regards to implementing the interval param. Are you saying that there's no value in doing that at all for this PR? If so, I don't really see how that implementation would change substantively as we go along seeing as its the implementation we need for this CLI to work as expected.

jpolchlo: I'm not 100% sure what you're getting at here.  I was putting up some incomplete work to foster discussion about what the output should look like.  We don't yet have an idea of what each interval's report should look like, which I think is the most important thing to figure out at this point.  I assume that there will be some kind of match/case that will process the grouped values differently based on the interval, but aside from that, we're lacking in the substance of those aggregations.  This initial commit was not a full and opinionated statement, just a rough structure so that we can stub out the remaining functionality.

In that vein, can we shift this discussion to what we want to see come out of the back end of this pipe—what data should be delivered to the subscriber for each interval?  Then I think we can fill in the provided armature.

CloudNiner: Yeah, the general stub here is certainly sufficient in terms of getting at some AugmentedDiffs and putting them in a DataFrame. No issues there. 

> about what the output should look like

Makes sense. For output, we need to end up with data in a structure that can be passed to a method that [sends emails via SES](https://docs.aws.amazon.com/ses/latest/DeveloperGuide/send-using-sdk-java.html). 

So maybe that's something like `List[Notification]` where `Notification` is something like:
```scala
// type aliases only to give meaning to base types
type Username = String
type EditCount = Int
case class Notification(userId: UUID, aoiName: String, start: DateTime, end: DateTime, geom: Geometry, counts: [(UserName, EditCount)])
```

???

If github comments are too low bandwidth, we can definitely chat offline. LMK.

CloudNiner: This was decided offline:

Seq[AugmentedDiff] + Seq[Feature[Geom, Data]] (from DB) --> Seq[email, aoiName, interval, Seq[(User, Info)]] where email, aoiName come from Data in DB and the rest is generated from this task.

CloudNiner:
Closing in favor of #161 so the parent feature branch is on the osmesa repo and can be more easily PRed to.

PR #162, opened 2019-10-15
Integrate postgres Notification list with AugmentedDiff source

CloudNiner:
The postgres table containing notifications was simplified such that we now just have a Notification table that directly contains each geometry and its name. This is because the MVP prototype specifically describes a notification as relating one to one to an AOI geom. No sense building something more complicated to prove a concept. We can always re-normalize later. Given that we want to just group on notification id. 

The data types were refactored a bit in order to match names to the types a bit better.

Deletes the remaining references/usages to the stubs that provided static geojson data.

### DIFF
--- a/src/analytics/src/main/scala/osmesa/analytics/oneoffs/StreamingAOIMonitor.scala
+++ b/src/analytics/src/main/scala/osmesa/analytics/oneoffs/StreamingAOIMonitor.scala
@@ -3,7 +3,6 @@ package osmesa.analytics.oneoffs
 import java.net.URI
 import java.sql.{Connection, DriverManager, Timestamp}
 import java.time.Instant
-import java.util.UUID
 
 import cats.data.Validated
 import cats.implicits._
@@ -72,11 +71,12 @@ object StreamingAOIMonitor
             implicit val spark: SparkSession = Analytics.sparkSession("Streaming AOI Monitor")
 
             import spark.implicits._
-            spark.withJTS
             import AOIMonitorUtils._
 
             val notifications: List[Notification] = queryNotifications(interval)
             notifications.foreach { println(_) }
+            val notificationsDf = notifications
+              .toDF("notificationId", "userId", "geom", "name", "email")
             val aoiIndex = AOIIndex(notifications)
 
             lazy val currentSequence = AugmentedDiffSource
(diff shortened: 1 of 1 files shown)


src/analytics/src/main/scala/osmesa/analytics/oneoffs/StreamingAOIMonitor.scala
```diff
@@ -136,42 +132,42 @@ object StreamingAOIMonitor
                 .map(s => Map(Source.EndSequence -> s.toString))
                 .getOrElse(Map.empty[String, String])
 
-            val aoiIndex = getAreaIndex( /*databaseUri*/ )
             val aoiTag = udf { g: jts.Geometry =>
               aoiIndex(g).toList
             }
+            val notificationId = udf { n: NotificationData =>
```
CloudNiner: We need to pull the notification id from the case class in order to group on it. Is there a cleaner way to do this that doesn't involve a UDF?

CloudNiner:
I still haven't been able to get this to successfully run in develop, even passing `AWS_PROFILE=<correct profile>` to sbt or setting it in my shell environment, I always get auth errors against the bucket it tries to pull augmented diffs from. Could you verify that this is an issue for you as well, or are you able to get this to run now?

src/analytics/src/main/scala/osmesa/analytics/oneoffs/StreamingAOIMonitor.scala
jpolchlo: Probably you can just `select` the `notificationId` field from `'data`

CloudNiner:
@jpolchlo this could use another look. With the new changes, I was able to get this to run to an exit 0 return code and display a match for my test area:

Screen Shot 2019-10-31 at 4 52 46 PM

src/analytics/src/main/scala/osmesa/analytics/oneoffs/StreamingAOIMonitor.scala
```diff
@@ -66,27 +59,30 @@ object StreamingAOIMonitor
...
-            spark.withJTS
             import AOIMonitorUtils._
 
-            val areasOfInterest: List[NotificationData] = queryAreasOfInterest(interval)
-            areasOfInterest.foreach { println(_) }
+            val notifications: List[Notification] = queryNotifications(interval)
+            notifications.foreach { println(_) }
+            val notificationsDf = notifications
+              .toDF("notificationId", "userId", "geom", "name", "email")
+            val aoiIndex = spark.sparkContext.broadcast(AOIIndex(notifications))
```
CloudNiner: Broadcasting AOIIndex was the solution for the PreparedGeom not serializable issue. This object is read-only, shared across the cluster, and should fit in memory, all the big prereqs for being a broadcast variable.

CloudNiner:
>Managed to compile and run to completion for the daily interval.

WHOOP. 🎉

PR #175, opened 2019-11-21
Update StreamingAOI notification to print change + edit counts per user

CloudNiner:
I had a miserable time emulating RDD.groupByKey in DataFrame. I'd be very curious if you have suggestions on how to avoid dropping to RDDs for this particular aggregation now that there's a working RDD implementation to compare against.

### DIFF
--- a/src/analytics/src/main/scala/osmesa/analytics/oneoffs/StreamingAOIMonitor.scala
+++ b/src/analytics/src/main/scala/osmesa/analytics/oneoffs/StreamingAOIMonitor.scala
@@ -74,7 +74,6 @@ object StreamingAOIMonitor
         (intervalOpt, augmentedDiffSourceOpt, startSequenceOpt, endSequenceOpt).mapN {
           (intervalArg, augmentedDiffSource, startSequence, endSequence) =>
             val appName = "StreamingAOIMonitor"
-            val environment = Properties.envOrElse("ENVIRONMENT", "development")
             val now = Timestamp.from(Instant.now)
 
             implicit val spark: SparkSession = Analytics.sparkSession("Streaming AOI Monitor")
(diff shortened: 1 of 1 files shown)


src/analytics/src/main/scala/osmesa/analytics/oneoffs/StreamingAOIMonitor.scala
```diff
@@ -247,30 +259,39 @@ object StreamingAOIMonitor
...
 object AOIMonitorUtils extends Logging {
 
-  case class NotificationData(notificationId: String, userId: String, name: String, email: String)
+  case class NotificationUserChangeSummary(notificationId: String,
+                                           user: String,
+                                           editCount: Long,
+                                           changesetCount: Int)
 
   // notificationId, userId, geom, name, email
   type Notification = (String, String, jts.Geometry, String, String)
```
jpolchlo: Make this a case class.  Improves readability, allows better interaction with Dataset/DataFrame APIs.

src/analytics/src/main/scala/osmesa/analytics/oneoffs/StreamingAOIMonitor.scala
```diff
@@ -93,8 +93,10 @@ object StreamingAOIMonitor
 
             val notifications: List[Notification] = queryNotifications(interval)
             notifications.foreach { println(_) }
-            val notificationsDf = notifications
-              .toDF("notificationId", "userId", "geom", "name", "email")
+            val notificationsRdd = spark.sparkContext.parallelize(notifications).map {
```
jpolchlo: Don't pack into an RDD.  Use `spark.createDataset(notifications)` (requires that `notifications` are filled with a case class instance).  You can then `groupByKey(_.notificationId)` or `join` using a more DataFrame-like interface.

CloudNiner: Updated. This is definitely cleaner and clearer.

src/analytics/src/main/scala/osmesa/analytics/oneoffs/StreamingAOIMonitor.scala
```diff
@@ -174,35 +171,54 @@ object StreamingAOIMonitor
 
             // 2. EXTRACT SALIENT INFO FROM DIFFS
             //    Prepare a dataset of summaries, one for each notification to send.
-            val messageInfo: Dataset[NotificationSummary] = {
+            val changeSummary = {
               val stats = diffs.withDelta
                 .withColumn("measurements", DefaultMeasurements)
                 .withColumn("counts", DefaultCounts)
```
jpolchlo: Any thoughts about either using this or killing it?

CloudNiner: Ah true. We're not using this at the moment. I suppose we just delete along with the commented out code and re-add later if that's what ends up being more useful.

PR #177, opened 2019-12-10
Split oneoffs to osmesa.apps package

CloudNiner:
These moves were done with IntelliJ "Move Class (F6)"
refactor tool. Using this tool does the following:
- Updates package reference
- Updates any import references (mostly automatic)
- Removes unused import references
- Reorders import references to alphabetical

I had to manually update most package/import refs
in docstring comments.

For now I just copy/pasted the analytics/build.sbt to apps/build.sbt. ~~Once everything is moved I'll go through and attempt to clean up the build.sbt's (dependencies, etc).~~ Edit: See comment below, this turned out to be a subpar experience. Taking suggestions for further improvements here if we want to address them now.

## Demo

Deployed the changes and observed the checkpoints table increment, and the changes to the class/jar names were applied in the ECS task definition.

Screen Shot 2019-12-11 at 9 36 04 AM
Screen Shot 2019-12-11 at 9 47 02 AM


src/apps/src/main/scala/osmesa/apps/batch/ChangesetStats.scala
```diff
@@ -1,4 +1,4 @@
-package osmesa.analytics.oneoffs
+package osmesa.apps.batch
```
jpolchlo: Is this a good candidate to create an `osmesa.apps.example` package for?  This utility doesn't fit into the workflow as such.  It only creates CSV output, though you could suggest that the CSVs could be used to ingest a  DB as well as the direct write operation.  One way or another, it's not part of our process anymore, so we either move this out to a less confusing place, or we rename to something less confusing, like `IngestToCSV`.

src/apps/src/main/scala/osmesa/apps/batch/ChangesetStatsCreator.scala
```diff
@@ -1,21 +1,21 @@
...
 import java.net.URI
-import java.sql._
 
 import cats.implicits._
 import com.monovore.decline.{CommandApp, Opts}
 import org.apache.spark.TaskContext
 import org.apache.spark.sql.SparkSession
 import org.apache.spark.sql.functions._
 import osmesa.analytics.Analytics
+import osmesa.analytics.oneoffs.MergeChangesetUtils
```
jpolchlo: Probably not a good import to use generally.  If there's useful functionality here, I recommend moving it to a less application-specific location (say to a utils pacakge object?).  (But is this import even used?)

src/apps/src/main/scala/osmesa/apps/streaming/AOIMonitor.scala
```diff
@@ -20,15 +20,15 @@ import org.locationtech.jts.{geom => jts}
 import org.locationtech.jts.geom.prep._
 import org.locationtech.jts.io.WKBReader
 import osmesa.analytics.Analytics
-import osmesa.analytics.oneoffs.Interval._
 import osmesa.analytics.stats._
+import osmesa.apps.streaming.Interval._
```
jpolchlo: If `Interval` has a broad application, I suggest maintaining it in the `analytics` package so that when we publish that package as a library, it will be available more broadly (for example, when this `AOIMonitor` class moves out to the SotD repo.

src/apps/src/main/scala/osmesa/apps/batch/ChangesetStats.scala
CloudNiner: Going to reach out to you on slack to address this one, there's a few potentially interconnected topics at play here that relate to what you've brought up.

src/apps/src/main/scala/osmesa/apps/batch/ChangesetStatsCreator.scala
CloudNiner: This is a rider object on MergeChangesets.scala which got moved to osmesa.apps.batch in the next commit I haven't yet pushed. So this Utils object was moved to osmesa.apps. 

I'll see about putting it somewhere more useful.

src/apps/src/main/scala/osmesa/apps/streaming/AOIMonitor.scala
CloudNiner: Nope. It currently doesn't. It's defined in this file and will move with the code when we move this command elsewhere since its just a duplicate of an object that SotD already defines. Just going to leave as is for now.

src/apps/src/main/scala/osmesa/apps/batch/ChangesetStats.scala
CloudNiner: Summary of followup:

This file is out of date, generating a db schema we no longer use and is therefore of extremely limited utility. We elected to just delete this file instead and remove its deployment ref from the batch Makefile.

src/analytics/build.sbt
```diff
@@ -29,7 +29,6 @@ libraryDependencies ++= Seq(
   gtSparkTestKit,
   logging,
   scalatest,
-  apacheCommonsEmail,
```
CloudNiner: I attempted to figure out if any of the other dependencies here are no longer necessary with a bunch of code moving to `osmesa.apps`. I was largely unsuccessful:

1. Use [sbt-explicit-dependencies](https://github.com/cb372/sbt-explicit-dependencies) plugin to auto find unused dependencies. This plugin told me that all my deps were unused, which isn't possible. There's a few issues where people hit similar problems.
2. Use [sbt-dependency-graph](https://github.com/jrudolph/sbt-dependency-graph) plugin to list dependencies, and then look for missing ones in the list. No go, the list is massive, doesn't print well and I can't limit only to top level deps. 
3. Manually remove deps one by one. Low confidence, due to lack of test suite and things compiling that I'm not sure should? e.g. I removed the `decline` dependency but that's required by code in the `osmesa.analytics.examples` module and I was still able to compile and generate an assembly jar.
4. Use IntelliJ's [dependency analyzer](https://www.jetbrains.com/help/idea/working-with-module-dependencies.html). This ran really slowly and gave similar results to (3) and there's no search in the list to quickly look up deps. So this would require a slow manual process I don't have time for.

As a result, I only removed the dep that I know is no longer necessary as it was just added in #174 and only used in the AOIMonitor command app that was moved here to `osmesa.apps`. Perhaps we can consider a more thorough cleanup of subproject deps and build config at a later time.

CloudNiner:
@jpolchlo this is ready for another look, I've completed the originally scoped set of changes and tested them successfully on the staging site (see Demo above). I'm in the process of reverting the staging site changes, as these ones will break a downstream app until I'm able to update the class path refs over there as well. 

I'm happy to attempt to address any other changes you had in mind now, as I have about 8 hours left based on the original estimate for this task, so theres some leeway _I think_ to do some additional work.

CloudNiner:
> README needs updating

🤦‍♂ good catch. Will tweak that and the other mentions momentarily.

> Also a question about whether the analytics/examples directory should be moved into the new apps package (osmesa.apps.examples).

Yeah, I debated this internally yesterday. I don't have a strong opinion so I decided to leave them where they are because they're examples of how one might use the constructs defined in the `analytics` project. They're good templates for writing your own command apps and I figured it would be good to have them closer to the source that people will be looking at. I'm happy to move them though given a constructive argument for putting them in `apps`. Thoughts from your end?

jpolchlo:
Ah, right.  Somehow it slipped my mind that we discussed moving these examples to Vectorpipe, since they don't stretch the OSMesa-provided features at all.  Let me suggest that they move there instead.  I think the other apps are demonstration enough of the feature set of OSMesa.

CloudNiner:
> moving these examples to Vectorpipe

👍 right. Going to address now since I have some time and this shouldn't take long.

PR #179, opened 2019-12-12
AOIMonitor command app miscellaneous cleanup

CloudNiner:
Fixes / cleans up a few issues in the AOIMonitor command app. See commit log for details on each fix.

The fallout of removing the unused withDelta call, along with adding a VectorPipe default spark session provider in geotrellis/vectorpipe#122 is that this command app no longer relies on anything specific to osmesa.


CloudNiner:
Going to merge this to get the issue resolved in staging, can address comments with later follow up.

PR #195, opened 2020-04-22
Add measurements + counts for landuse and natural tags

CloudNiner:
For measurements, only include landuse and natural objects
that are areas.

For counts, count everything.

I switched `withDelta` to `withLinearDelta` to clarify
that it only computes deltas for linear features. An
alternative implementation would be to have the
original `withDelta` method compute both area and linear
deltas, but then its unclear in the generated column
whether the number has length or area units.

VectorPipe uses an outdated set of area keys from
github.com/osmlab/id-area-keys. Counts _should_ be more
accurate if we update to the latest area keys list. @mojodna 
💯 points for the comment noting where the area keys map
came from including version! 
Here's the current diff:
osmlab/id-area-keys@f645eca#diff-f36fb8b1baaccf032007cfcc70836c1d

### DIFF
--- a/src/analytics/src/main/scala/osmesa/analytics/stats/package.scala
+++ b/src/analytics/src/main/scala/osmesa/analytics/stats/package.scala
@@ -1,16 +1,26 @@
 package osmesa.analytics
 
+import geotrellis.proj4.{CRS, LatLng, WebMercator}
+import geotrellis.proj4.util.UTM
+import geotrellis.vector.{Geometry => GTGeometry}
 import org.apache.spark.sql.expressions.Window
 import org.apache.spark.sql.{Column, DataFrame, SparkSession}
 import org.apache.spark.sql.types.IntegerType
 import org.apache.spark.sql.functions._
+import org.locationtech.geomesa.spark.jts.{udf => _, _}
+import org.locationtech.jts.geom.Geometry
 import osmesa.analytics.stats.functions._
 import vectorpipe.functions.osm._
-import org.locationtech.geomesa.spark.jts._
 
 package object stats {
 
-  @deprecated()
+  val transformLatLngToUtm = udf { g: Geometry =>
+    val centroid = g.getCentroid
+    val utmCrs = UTM.getZoneCrs(centroid.getX, centroid.getY)
+    GTGeometry(g).reproject(LatLng, utmCrs).jtsGeom
+  }
+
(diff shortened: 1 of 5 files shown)


src/analytics/src/main/scala/osmesa/analytics/stats/package.scala
```diff
@@ -64,27 +89,34 @@ package object stats {
   def isOther(tags: Column): Column = isTagged(tags) and
     not(isRoad((tags))) and
     not(isWaterway(tags)) and
-    not(isCoastline(tags)) and
```
CloudNiner, line 67: Removed because isCoastline is included by the new isNatural check below.

src/analytics/src/main/scala/osmesa/analytics/stats/package.scala
```diff
@@ -46,6 +65,12 @@ package object stats {
     isCoastline(tags) or
     isPOI(tags) as 'isInterestingWay
 
+  def isLanduse(tags: Column): Column =
+    tags.getItem("landuse").isNotNull as 'isLanduse
+
+  def isNatural(tags: Column): Column =
+    tags.getItem("natural").isNotNull as 'isNatural
+
```
jpolchlo, line 73: Is there a problem with double counting coastlines in the natural category?  How about railways in landuse?  I doubt the separate category should be removed, but should the more catchall categories have those tags omitted?

CloudNiner: I asked the client about this. They're ok with double counting these cases.

CloudNiner: Good catch though.

CloudNiner:
Tested on client staging environment. Here's an excerpt from one of the changesets I used to test:
```
id              | 954222
measurements    | {"road_km_modified": 0.0000000007130242920538876, "landuse_km2_added": 1.8527665672200757, "natural_km2_added": 3.9711439174635026}
counts          | {"other_added": 1, "landuse_added": 2, "natural_added": 1, "roads_modified": 2}
total_edits     | 6
```

It looks like performance creating the batch backfill is approximately the same. Previous runs took ~40 mins with the config in this repo, now they take maybe a few minutes more at most.

@jpolchlo this change is worth another look, there's been a few tweaks since you last looked.

PR #198, opened 2020-07-24
Add foreign keys and indices to postgresql schema.

guidorice:
This PR modifies the PG database schema.

- Makes foreign keys explicit.
- Adds indices to support joins with foreign keys, and to support queries we have seen in production.

### Known issues

- Some tables may contain so few records that the PG query planner will not actually  use the indices, but at least they will be in place for future.
- Adding the constraint `changesets_users_id_fk` into `changesets` is going to cause problems: Executing this on a development db, I got an error because some user_ids are not present in the user table:

```sql
alter table changesets
add constraint changesets_users_id_fk
foreign key (user_id) references users;
```

The missing users issue should be resolved before changing the table create statement here.

Please beware because I have not done any integration testing and do not really have any way to see what side effects it will cause.  😄

@CloudNiner I marked this as a Draft PR, because of the user_id concern.

### DIFF
--- a/deployment/sql/changesets.sql
+++ b/deployment/sql/changesets.sql
@@ -4,7 +4,9 @@ CREATE TABLE changesets (
     counts jsonb,
     total_edits integer,
     editor text,
-    user_id integer,
+    user_id integer
+        CONSTRAINT changesets_users_id_fk
+		REFERENCES users,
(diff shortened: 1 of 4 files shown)


guidorice:
@CloudNiner  thanks for the assist- this is ready for review now.

PR #200, opened 2020-08-12
Facilities for bulk ingest file prep

jpolchlo:
When reconstructing the staging DB, it is necessary to produce an ORC file for the planet history as well as to update a changeset ORC file. The latter was covered by a make rule, but the former task was previously left to a manual process. This PR adds automation to help with this, and combines it with the changeset ORC update. Now, from `./deployment/batch/makefile`, `make prepare-bulk-ingest` will create a cluster and add steps for both the planet history ORC conversion and the changeset ORC update. The cluster will terminate automatically after the steps terminate.


jpolchlo:
Sorry.  Commit log is a mess from a mistaken addition of some verboten text and the subsequent history rewrite.

jpolchlo:
Closing in favor of #206 and it's clean tree.

PR #201, opened 2020-08-13
Introduce infrastructure for augdiff publisher

jpolchlo:
We hadn't included automation infrastructure for the augmented diff publisher, meaning that the process for setting it up was basically undocumented.  This PR goes some ways toward fixing that.  We still require that a container built from https://github.com/mojodna/overpass-diff-publisher be provided manually, but the task definitions are then provided by the makefiles given here.


deployment/streaming/config-deployment.mk.template
```diff
@@ -9,15 +9,20 @@ export IAM_ACCOUNT :=
...
 #                                                 Streaming resource definitions
 ################################################################################
-export CLUSTER_NAME := osm-stat-stream-cluster
 export STREAMING_INSTANCE_TYPE := m4.xlarge
 export ECR_IMAGE :=
 export AWS_LOG_GROUP := streaming-stats-updater
 export ECS_SUBNET := ${SUBNET}
 export ECS_SECURITY_GROUP :=
 
+export CLUSTER_NAME_DEPLOYMENT :=
```
CloudNiner, line 18: What are these two vars used for? They're not elsewhere in the diff -- can they be removed?

deployment/streaming/config-deployment.mk.template
```diff
@@ -28,9 +33,12 @@ export NETWORK_CONFIGURATION="{\"awsvpcConfiguration\": {\"subnets\": [\"${ECS_S
 ################################################################################
 #                                                     Batch resource definitions
 ################################################################################
-export SERVICE_ACCESS_SECURITY_GROUP := ${ECS_SECURITY_GROUP}
-export EMR_MASTER_SECURITY_GROUP :=
-export EMR_SLAVE_SECURITY_GROUP :=
+export MASTER_SECURITY_GROUP :=
```
CloudNiner, line 36: Sort of the same question here, looks like a rename. Just double checking that this is fixing up a var name to the correct value rather than accidentally introducing an error.

jpolchlo:
Replacing this PR with #207 to get a cleaner commit history

PR #207, opened 2020-09-14
Introduce infrastructure for augdiff publisher

jpolchlo:
We hadn't included automation infrastructure for the augmented diff publisher, meaning that the process for setting it up was basically undocumented.  This PR goes some ways toward fixing that.  We still require that a container built from https://github.com/mojodna/overpass-diff-publisher be provided manually, but the task definitions are then provided by the makefiles given here.

This PR replaces #201 after the latter was polluted with private information.


CloudNiner (approved):
Note that the container should now be built from https://github.com/azavea/overpass-diff-publisher to include multi polygon relation information, rather than https://github.com/mojodna/overpass-diff-publisher as indicated in the PR description. That doesn't change anything about this PR though.

PR #209, opened 2020-11-03
Changeset metadata ORC improvements

jpolchlo:
This PR adds some infrastructure to the batch makefiles to allow for the easy creation of a changeset metadata ORC from CSV exports from an OSM database.


CloudNiner (approved):
👍 Don't forget to update our saved config-run.mk with any changes (looks like the config-deployment.mk is already up to date)

PR #211, opened 2020-11-20
Update augdiff tasks to match deployment

CloudNiner:
- Now use the correct azavea-overpass-diff-publisher
  task definitions (since we switched to azavea fork
  of the repo)
- Support deployments using Onramp

I tested this in our staging and production environment by setting the appropriate config-deployment.mk variables and running the define then deploy tasks.


jpolchlo (approved):
LGTM

PR #212, opened 2020-11-25
Consolidate makefiles

### LINKED ISSUE #189: Migrate batch scripts in deployment/streaming to use batch-process.json
Currently batch-generate-edit-histograms.sh and batch-generate-footprints.sh have separate scripts from generate db backfill with uses batch-process.sh. With the addition of this more generic batch-process.sh we could just have separate json configurations for the instance/spark conf requirements for these processes and not maintain almost duplicate shell scripts. 

It should be noted that as of writing, batch-process takes its instance type configuration via env var since the other scripts do as well. If this task is addressed, it likely makes sense to migrate this env var into a parameter on the batch-process.sh script.

### LINKED ISSUE #203: Migrate materialized views from osmesa-stat-server
Now that the stat server is deprecated, we should import the materialized views and associated `make` targets into the osmesa repository.  This will allow our deployments to be performed entirely from this repository, and we can fully abandon the stat server.

jpolchlo:
The purpose of this PR is manifold, but summed up as "simplify the deployment process".  The aim is to consolidate as many steps as possible into make rules that can be executed reliably, without fuss.

Knock on benefits of this PR include the ability to finally retire the osmesa-stats-server by importing the necessary materialized view definitions and refresh infrastructure.  This PR will also automate the process of rebuilding the changeset metadata from OSM DB extracts.  This will also take a step towards eliminating the `deployment/batch` subdirectory, and ultimately allowing us to eliminate the distinction between batch and streaming deployments (though the batch infrastructure still provides the means to create EMR clusters for direct interaction (i.e., not pre-canned operations)). 

Closes #203 
Closes #189


deployment/streaming/scripts/define-staging-view-refresher.sh
```diff
@@ -0,0 +1,38 @@
...
+                \"options\": {
+	          \"awslogs-group\": \"/ecs/${AWS_LOG_GROUP}-staging\",
+	          \"awslogs-region\": \"${AWS_REGION}\",
+                  \"awslogs-stream-prefix\": \"ecs\"
+                }
+              },
+              \"command\": [
+                \"refresh-views.sh\"
+              ],
+              \"environment\": [
```
CloudNiner, line 29: Need to add additional env variable here: `DATABASE_NAME` set to `STAGING_DB`. Without it, the refresher tasks don't correctly check if a matching task is already executing and begin to pile up.

CloudNiner (changes requested):
Only one immediate comment for now, since it breaks the staging env. 

More complete review tomorrow...

deployment/Dockerfile
```diff
@@ -1,5 +1,20 @@
...
+
+RUN apt-get dist-upgrade -y
+
+RUN \
+  apt-get install -y --no-install-recommends postgresql-client \
+  && rm -rf /var/lib/apt/lists/*
+
 COPY osmesa-apps.jar /opt/osmesa-apps.jar
 COPY streaming/log4j.properties /spark/conf/
+COPY refresh-views.sh /usr/local/bin/refresh-views.sh
```
CloudNiner, line 5: It looks like with these changes that we've got one container image pulling double duty for two entirely separate tasks. 

As far as I can tell, the image running refresh-views could just be a very small base alpine linux image with psql package installed. If we moved it to a separate image we'd also presumably get faster start times when we trigger these tasks since the AWS executor would be downloading much smaller images to run. 

Can you create a follow up issue to split this Dockerfile in two?

deployment/streaming/Makefile
```diff
@@ -25,6 +27,18 @@ BATCH_WORKERS=1
 endif
 endif
 
+ifndef TURBO_BOOST
```
CloudNiner: 🚀

deployment/streaming/Makefile
```diff
@@ -34,7 +48,19 @@ endif
 
 .PHONY: build-container login-aws-registry tag-image push-image
 
-build-container:
+APPS_SRC_DIR := ../../src/apps
+APPS_ASSEMBLY := ${APPS_SRC_DIR}/target/scala-2.11/${OSMESA_APPS_JAR}
+
+rwildcard=$(foreach d,$(wildcard $1*),$(call rwildcard,$d/,$2) $(filter $(subst *,%,$2),$d))
```
CloudNiner, line 41: 👀 what does this do?

deployment/refresh-views.sh
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env bash
```
CloudNiner: Would it make sense to invert the structure of this folder a bit? 

Current:
- deployment/refresh-views.sh
- deployment/sql/materialized-views/*.sql

Proposed:
- deployment/materialized_views/refresh-views.sh
- deployment/materialized_views/sql/*.sql

This keeps folders under `deployment/` loosely organized by topic, e.g. batch, streaming, materialized_views and keeps a script (refresh-views) intended to be run on infrastructure, out of the root deployment directory which currently contains artifacts and scripts to be run locally that build said infrastructure.

EDIT: Looking at the dir, it appears we also have create table statements as well. Perhaps those could be moved to `deployment/db-init/*.sql`. Or, this is now too complicated and disregard this comment 😄

deployment/streaming/scripts/batch-process.sh
```diff
@@ -79,6 +79,8 @@ aws emr create-cluster \
         }
       }
     ]" \
+  --scale-down-behavior TERMINATE_AT_TASK_COMPLETION \
```
CloudNiner, line 82: 👍

deployment/streaming/scripts/define-production-streaming-update-tasks.sh
```diff
@@ -37,7 +37,7 @@ aws ecs register-task-definition \
 	        }
 	      ],
 	      \"image\": \"${ECR_IMAGE}:production\",
-	      \"name\": \"streaming-augmented-diffs-stats-updater\"
+	      \"name\": \"streaming-changeset-stats-updater\"
```
CloudNiner, line 40: 👍

CloudNiner (approved):
⭐ this is a really nice iterative improvement to simplify the user experience of deployments, even if we've added a few new variables to consider.

deployment/streaming/Makefile
CloudNiner: I've had a conversation with ops on the client side regarding this over the past week. They are comfortable with us maintaining the higher resource allocations permanently since they plan to do more regular large bulk imports.

As a result, we should just remove the TURBO_BOOST if else but leave the variables in to be configured by the user. In the case of our deployment we want them to be configured at the new higher boosted values:
```
NODE_OPTIONS=--max-old-space-size=7168
DRIVER_MEMORY=8192m
ECS_MEMORY=8
AUGDIFF_ECS_MEMORY=8
```

deployment/streaming/scripts/define-staging-streaming-update-tasks.sh
```diff
@@ -12,7 +12,7 @@ aws ecs register-task-definition \
     --network-mode awsvpc \
     --requires-compatibilities EC2 FARGATE \
     --cpu "1 vCPU" \
-    --memory "4 GB" \
+    --memory "${ECS_MEMORY} GB" \
```
CloudNiner: I'd make one of two changes focused on improving clarity:

1. Rename `ECS_MEMORY` to `ECS_MEMORY_GB`
2. Include the "multiplier" e.g. GB in the variable value, so that the user can specify in kb,mb,gb whatever works for them. This would be consistent with our use of other memory variables in this project, like `DRIVER_MEMORY`.

deployment/streaming/scripts/define-streaming-augdiff-producer.sh
```diff
@@ -12,7 +12,7 @@ aws ecs register-task-definition \
     --network-mode awsvpc \
     --requires-compatibilities EC2 FARGATE \
     --cpu "1 vCPU" \
-    --memory "2 GB" \
+    --memory "${AUGDIFF_ECS_MEMORY} GB" \
```
CloudNiner: I'd make one of two changes focused on improving clarity:

1. Rename `AUGDIFF_ECS_MEMORY` to `AUGDIFF_ECS_MEMORY_GB`
2. Include the "multiplier" e.g. GB in the variable value, so that the user can specify in kb,mb,gb whatever works for them. This would be consistent with our use of other memory variables in this project, like `DRIVER_MEMORY`. 

Let's make the same change for both this var and `ECS_MEMORY`.

deployment/streaming/Makefile
jpolchlo: Neat trick to allow us to check for changes to `.scala` files so that we don't needlessly rebuild the assembly.  This gets used like a function (the $1 and $2 are the arguments).  See the invocation in the rule just below.  Thanks to smarter people than me for figuring this out.

deployment/refresh-views.sh
jpolchlo: In order to address the mess of docker build resources, I pulled `refresh-views.sh` into a new directory.  That sorta makes it unnecessary to make a `materialized-views` directory.  I think the `sql` directory name is pretty clear?  There's a README in that directory that should clear up confusion.  Happy to revisit if needed.

jpolchlo:
The Docker resources have been restructured.  There are now two images for the two distinct tasks, and the structure has been verified in use.  This is better, since adding `psql` to the spark container was a mess, since that image is now on an out of date debian release for which there are no longer apt profiles.  Builds are faster and more maintainable for when we bump to Spark 3.

deployment/docker/refresh-views.sh
```diff
@@ -1,50 +1,50 @@
 #!/usr/bin/env bash
 
-echo "$(date --iso-8601=seconds): Starting view refreshment in $DATABASE_NAME"
+echo "$(date -Iseconds): Starting view refreshment in $DATABASE_NAME"
```
CloudNiner, line 3: This was changed because the invocation is different now that  it's running in `alpine`, right?

CloudNiner (approved):
🔥

deployment/refresh-views.sh
CloudNiner: Updated changes are great.

deployment/docker/refresh-views.sh
jpolchlo: Yeah, evidently the `--iso-8601` option is not standard.  The `-I` is (supported by both GNU and non-GNU `date` commands).

PR #213, opened 2020-12-15
Add script to send email when checkpoints fall behind

CloudNiner:
See script header for usage docs. This script is a quick way for us to get alerts independent of the OSMesa pipeline when the streaming updaters fall behind.

The intent is to add this script to CloudWatch Events and run it on a regular schedule, perhaps every five minutes. I'll open a separate PR with the infrastructure component once #212 is merged.

@jpolchlo when you test this, shoot me a slack message and I'll pass along testing instructions on our staging environment.

# Demo

Screen Shot 2020-12-15 at 4 42 26 PM


jpolchlo (approved):
Can confirm that the script given here sends an email.

